PR #5, opened 2016-07-09
Fix failing TestRecordReaderFunction and TestRecordReaderBytesFunction

### LINKED ISSUE #2: Spark tests failing
Datavec spark:
- TestRecordReaderBytesFunction.testRecordReaderBytesFunction
- TestRecordReaderFunction.testRecordReaderFunction

I suspect that this might be due to ParentPathLabelGenerator having difficulty when used with spark, but I'm not 100% sure there.

saudet:
Fixes #2 

@AlexDBlack Or would there be a better way of doing this than adding a `setLabels()` method?


AlexDBlack:
Look good to me. I can't think of a better way to do this, for Spark.

PR #18, opened 2016-07-16
Replicate cnaova PR https://github.com/deeplearning4j/Canova/pull/230

agibsonccc:
https://github.com/deeplearning4j/Canova/pull/230


datavec-api/src/main/java/org/datavec/api/records/reader/impl/LineRecordReader.java
```diff
@@ -154,7 +159,6 @@ public void reset() {
         if(inputSplit == null) throw new UnsupportedOperationException("Cannot reset without first initializing");
         try{
             initialize(inputSplit);
-            currIndex = 0;
```
akhodakivskiy: Hey @agibsonccc  did you somehow override this change?

PR #26, opened 2016-07-22
fix transitive scala versions

agibsonccc:


saudet:
Looks good to me. It'd be hard to do better than that with Maven :)

BTW, there is a convention for the ordering of elements in the pom.xml file:
https://maven.apache.org/developers/conventions/code.html

PR #40, opened 2016-07-30
Resize transform

raver119:
Resize Image Transform


saudet:
Looks fine, just not sure what the use case for that would be...?

raver119:
Well, i just needed it because i'm trying to reproduce reported issue, and don't have properly preprocessed dataset. So with this transform i'm at least sure all images are equal sized :)

PR #43, opened 2016-08-02
Remove CLI

saudet:
Please merge if this looks alright...


AlexDBlack:
Looks OK to me. But want to double check with @jpatanooga first just in case he has code still using any of this.

jpatanooga:
its fine. as long as we replicate all of the vectorization functionality in
DataVec, we'll be ok

PR #56, opened 2016-09-05
Add color conversion transform

nyghtowl:


datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/ColorConversion.java
```diff
@@ -0,0 +1,79 @@
...
+        }
+        Mat mat = (Mat) converter.convert(image.getFrame());
+        Mat result = new Mat();
+
+        try {
+            cvtColor(mat, result, conversionCode);
+        }catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+        return image;
```
saudet: We want to return the result here:
```java
return new ImageWritable(converter.convert(result));
```

saudet:
The rest looks good! It's not using random anywhere, but that's probably alright, not sure how this could be made to use random...

PR #70, opened 2016-10-07
Improve Text Handling

### LINKED ISSUE #69: TfidfRecordReader does not reset
While trying to replace my own ETL pipeline, I found that the TfidfRecordReader doesn't reset when calling `.reset()` on it.

treo:
This Pull Request addresses 4 problems with the TfidfRecordReader:
* It allows TfidfRecordReader to be reset in order to be able to iterate multiple times over the RecordReaderDataSetIterator that is created from it
* It corrects the tfidf calculation by fixing a bug where tokens where counted multiple times per document
* It improves performance by moving only complete tfidf arrays over to ND4J instead of running `.putScalar` for each element
* It actually respects StopWords for tfidf

And it improves some other areas around vectorizing text with datavec
* It actually uses the preprocessor set on UimaTokenizer
* It allows to configure a custom VocabCache, e.g. if you want to implement a version that also has a MAX_WORD_FREQUENCY


agibsonccc:
Are you done on this?

treo:
Yep, this is done.

PR #107, opened 2016-11-11
add beginning of datavec local

agibsonccc:
Add table/record interop and a new data vec local module.


huitseeker:
[Datavec - add local executor](https://trello.com/c/QFy9g3E1/6-datavec-add-local-executor)

PR #131, opened 2016-11-30
Support for Spark 2.0 FlatMap incompatible changes

elyast:
Switching implementations from spark FlatMapFunction to internal one due to incompatibilities between Spark releases


elyast:
I investigated split of 2 projects into 3 base, spark1, spark2, however there is a problem with that approach, since base should have some implementation of flat map functions that its using, if we provide dummy implementation in common, then we may face class loading problems later on, since its not easy to remove dummy implementation, unless we remove them from jar after compilation.

Please advice if that what you want to do.

structure could be following:
```
/DataVec
 - datavec-spark
     - common (have same implementation as in spark1, for testing purposes, but is removed from including into jar), spark dependency will be provided, not to be carried over
     - spark1 (just flat map functions implementation same as in common) dependency to spark 1.6 and common
     - spark2 (just flat map functions implementation, different because interfaces were changed) dependency to spark 2.0 and common
```

Drawback of this approach is that in the future common should have duplicated implementations from spark1 e.g., I'd rather stick with my current implementation, using just 2 projects were spark1 has common code and delta, and spark2 just delta and is copying over spark1 classes, so both are self sufficient packages with proper implementations

Let me know what do you think.

agibsonccc:
This hasn't gone un noticed, @huitseeker is helping us decide how to do the spark 2.0 integrations.

elyast:
@huitseeker any progress on this topic ?

agibsonccc:
@elyast spark 2.0 just when GA in CDH yesterday. I will be looking at this over this week.

huitseeker:
Thanks for this PR !
However, I'm not sure about the `datavec-spark2` module and its place as an import which reflects only the changes for a specific Scala / Spark version.

Since we are only changing the source here to respect the [Spark API changes](http://www.spark.tc/migrating-applications-to-apache-spark-2-0-2/), it seems we could get away with using a source directory, that uses distinct source code depending on the Spark verison that is in the build. 

To do this I would suggest using the [build helper maven plugin](http://www.mojohaus.org/build-helper-maven-plugin/). It would look like this in the datavec-spark POM:

```
      <!-- added source folder containing the code specific to the spark version -->
      <plugin>
        <groupId>org.codehaus.mojo</groupId>
        <artifactId>build-helper-maven-plugin</artifactId>
        <version>1.12</version>
        <executions>
          <execution>
            <phase>generate-sources</phase>
            <goals><goal>add-source</goal></goals>
            <configuration>
              <sources>
                <source>src-spark-${spark.version}</source>
              </sources>
            </configuration>
          </execution>
        </executions>
      </plugin>
```

Then we could keep the same common classes (thanks for formulating an adapter they can all refer to !) in the `src` directory, and only defer to the spark-version-specific folder when needed. All of this in a single module.

datavec-spark2/pom.xml
```diff
@@ -0,0 +1,102 @@
...
+  ~  *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~  *    See the License for the specific language governing permissions and
+  ~  *    limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <artifactId>datavec-parent</artifactId>
+        <groupId>org.datavec</groupId>
+        <version>0.7.1-SNAPSHOT</version>
```
huitseeker: You're gonna have to update this version with the latest, even though this PR merges cleanly.

elyast:
@huitseeker That makes sense I will update my PR as soon as I get a chance

elyast:
@huitseeker please take a look, I also found out that DataFrame class doesn't exist anymore in spark 2 so I needed to copy DataFrames and Normalization, I suspect there could be better way of handling it

huitseeker:
@elyast You can see a pretty complete migration guide here:
http://www.spark.tc/migrating-applications-to-apache-spark-2-0-2/ It will help you not get surprised by things like the `DataFrame` "disparition"

elyast:
@huitseeker thanks for the link, let me know if there is anything I can do more, or this patch is good to go

datavec-spark/src/main/spark-1/org/datavec/spark/transform/join/ExecuteJoinFlatMapFunction.java
```diff
@@ -0,0 +1,43 @@
...
+import scala.Tuple2;
+
+import java.util.List;
+
+/**
+ * Execute a join
+ *
+ * @author Alex Black
+ */
+public class ExecuteJoinFlatMapFunction implements FlatMapFunction<Tuple2<List<Writable>,Iterable<JoinValue>>, List<Writable>> {
```
huitseeker: This class and `FilterAndFlattenJoinValues` fundamentally do the same thing. Could this be a generic class that takes a `FlatMapFunctionAdapter` ? The only thing that would change a bit is the constructor (to which you would need to pass the built instance) but it would get us a big reduction in duplication. 

I do admit this kind-of-verbose architecture is not inherently part of your Spark PR, but the duplication it causes kind of mandates the refactoring. I believe it has the potential to make things simpler too. (summoning @AlexDBlack for encouraging comments).

datavec-spark/src/main/spark-1/org/datavec/spark/transform/sparkfunction/SequenceToRows.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.UUID;
+
+/**
+ * Convert a record to a row
+ * @author Adam Gibson
+ */
+public class SequenceToRows implements FlatMapFunction<List<List<Writable>>,Row> {
```
huitseeker: Same remark as [this comment](https://github.com/deeplearning4j/DataVec/pull/131/files#r96178336)

datavec-spark/src/test/java/org/datavec/spark/transform/NormalizationTests.java
```diff
@@ -1,7 +1,7 @@
 package org.datavec.spark.transform;
 
 import org.apache.spark.api.java.JavaRDD;
-import org.apache.spark.sql.DataFrame;
+import org.apache.spark.sql.Dataset;
```
huitseeker: Aren't you assuming tests run on Spark 2.0 only? I'm afraid we might have to split test source based on Spark version as well.

datavec-spark/src/main/spark-2/org/datavec/spark/transform/DataFrames.java
```diff
@@ -0,0 +1,423 @@
...
+     * - Column 0: Sequence UUID (name: {@link #SEQUENCE_UUID_COLUMN}) - a UUID for the original sequence<br>
+     * - Column 1: Sequence index (name: {@link #SEQUENCE_INDEX_COLUMN} - an index (integer, starting at 0) for the position
+     * of this record in the original time series.<br>
+     * These two columns are required if the data is to be converted back into a sequence at a later point, for example
+     * using {@link #toRecordsSequence(Dataset<Row>)}
+     *
+     * @param schema Schema to convert
+     * @return StructType for the schema
+     */
+    public static StructType fromSchemaSequence(Schema schema) {
```
huitseeker: I'll buy the need to split a version of `DataFrames` and `Normalization`, but I do believe in the utility of minimizing the split / duplicated code absolutely. Is there a way we could have the members we absolutely need to split in an abstract class `DataFramesTemplate`, and members such as this one be in a usable class that extends it ?

datavec-spark/src/main/spark-2/org/datavec/spark/transform/Normalization.java
```diff
@@ -0,0 +1,400 @@
...
+    /**
+     * Normalize each column of a sequence, based on min/max
+     *
+     * @param schema Schema of the data
+     * @param data   Data to normalize
+     * @param min    New minimum value
+     * @param max    New maximum value
+     * @return Normalized data
+     */
+    public static JavaRDD<List<List<Writable>>> normalizeSequence(Schema schema, JavaRDD<List<List<Writable>>> data, double min, double max) {
```
huitseeker: Same as [there](https://github.com/deeplearning4j/DataVec/pull/131/files#r96179101)

pom.xml
```diff
@@ -72,6 +72,8 @@
         <freemarker.version>2.3.23</freemarker.version>
         <logback.version>1.1.2</logback.version>
         <maven-deploy-plugin.version>2.8.2</maven-deploy-plugin.version>
+        <spark.version>1.6.2</spark.version>
```
huitseeker: 👍

datavec-spark/src/main/spark-1/org/datavec/spark/transform/join/ExecuteJoinFlatMapFunction.java
elyast: I can provide abstract class that will unify common code, however FilterAndFlattenJoinValues and ExecuteJoinFlatMapFunction are public APIs, so they need to stay imho.

huitseeker: Fair enough.

datavec-spark/src/main/spark-2/org/datavec/spark/transform/DataFrames.java
elyast: Only problem I see is there are all static methods, as far as I checked there is a way to get couple of methods out of DataFrames, however in case of Normalization there is no chance

elyast:
@huitseeker I did a bit refactoring, I believe FlatMapFunction implementations looks good, there is still a lot copy-paste in Normalization and DataFrames, I believe if we would like to fix it, we would need to change static methods to member ones and utilize inheritance.

datavec-spark/src/main/java/org/datavec/spark/transform/analysis/SequenceFlatMapFunction.java
```diff
@@ -27,13 +27,10 @@
...
  */
-public class SequenceFlatMapFunction implements FlatMapFunction<List<List<Writable>>, List<Writable>> {
+public class SequenceFlatMapFunction extends BaseFlatMapFunctionAdaptee<List<List<Writable>>, List<Writable>> {
 
-    private final SequenceFlatMapFunctionAdapter adapter = new SequenceFlatMapFunctionAdapter();
-
-    @Override
-    public Iterable<List<Writable>> call(List<List<Writable>> collections) throws Exception {
-        return adapter.call(collections);
+    public SequenceFlatMapFunction() {
```
huitseeker: 👍 Much better !

datavec-spark/pom.xml
```diff
@@ -42,8 +43,18 @@
                 </sources>
               </configuration>
             </execution>
+            <execution>
+              <id>add-test-source</id>
```
huitseeker: 👍 thanks !

datavec-spark/src/main/spark-1/org/datavec/spark/transform/BaseFlatMapFunctionAdaptee.java
```diff
@@ -0,0 +1,17 @@
+package org.datavec.spark.transform;
+
+import org.apache.spark.api.java.function.FlatMapFunction;
+import org.datavec.spark.functions.FlatMapFunctionAdapter;
+
+public class BaseFlatMapFunctionAdaptee<K, V> implements FlatMapFunction<K,V> {
```
huitseeker: 👍 Nice!

datavec-spark/src/test/spark-1/org/datavec/spark/transform/DataFrameTests.java
```diff
@@ -0,0 +1,93 @@
...
+        for(int i = 0; i < arr.rows(); i++)
+            arr.getRow(i).addi(i);
+
+        List<List<Writable>> records = RecordConverter.toRecords(arr);
+        Schema.Builder builder = new Schema.Builder();
+        int numColumns = 10;
+        for(int i = 0; i < numColumns; i++)
+            builder.addColumnDouble(String.valueOf(i));
+        Schema schema = builder.build();
+        DataFrame dataFrame = DataFrames.toDataFrame(schema,sc.parallelize(records));
```
huitseeker: Would it be possible to hide the difference between this `DataFrame` and `DataSet<Row>` with a `spark-$sparkMajorVersion` class `DataRowsFacade` class that would be duplicated, and reformulate this using the facade ? I know it's annoying to have to modularize all this, but the maintenance of these tests will be daunting without this.

datavec-spark/src/test/spark-1/org/datavec/spark/transform/NormalizationTests.java
```diff
@@ -0,0 +1,118 @@
...
+                record.add(new DoubleWritable(1.0));
+            }
+
+        }
+
+        INDArray arr = RecordConverter.toMatrix(data);
+
+        Schema schema = builder.build();
+        JavaRDD<List<Writable>> rdd = sc.parallelize(data);
+        DataFrame dataFrame = DataFrames.toDataFrame(schema,rdd);
```
huitseeker: Same as  [this comment above](https://github.com/deeplearning4j/DataVec/pull/131/commits/23cb8b774ab8c64da48e3ce639ddec0ad6b74eab#r96313974), *mutatis mutandis*.

AlexDBlack:
I finally had the time to look at this, thanks again for driving this @elyast - overall it looks good to me.

One question here - when it comes to actually building and releasing both spark 1 and spark 2 variants: Is the approach here to simply change the spark versions (and spark major version), either manually, using profiles etc or something like this: https://github.com/deeplearning4j/DataVec/blob/master/buildmultiplescalaversions.sh

i.e., so a mvn install would do spark 1 (scala 2.10 and 2.11), change spark versions to 2, then we'd do a mvn install again (also with scala 2.10 and 2.11)?

huitseeker:
@AlexDBlack Good thinking ! The advantage of the build helper maven plugin, as outlined [here](https://github.com/deeplearning4j/DataVec/pull/131#issuecomment-269737318), is that source is just dependent on spark-version. So, by just using a `change-spark-version` script you would be switching, building on, and releasing for both spark variants.

datavec-spark/src/test/spark-1/org/datavec/spark/transform/DataFrameTests.java
elyast: @AlexDBlack @huitseeker 
We could also use DataRowsFacade in the code as well, however that would change API, e.g. instead of DataFrames.toDataFrame returning DataFrame it will return DataRowsFacade. 
What do you think?

elyast:
@huitseeker please take a look now, it looks pretty promising now

elyast:
I'd appreciate merging if everyone approves, I'd like to do similar exercise for deeplearning4j lib as well, however would be good to use base classes from DataVec

AlexDBlack:
I had a look through this, and ran through a build and the unit tests. Looks good to me. 👍 
I'm happy for this to be merged. We will need to update the build scripts here soon after doing so, but that can be done in a separate PR.

Also for reference in case it comes up again: I had some minor issues with intellij when switching spark versions, solved exclusions and marking/unmarking spark 1/2 sources as appropriate.

elyast:
@AlexDBlack here is a pull request for build scripts: https://github.com/deeplearning4j/DataVec/pull/171

PR #141, opened 2016-12-17
Support just the classify use case for writing a data set to a list o…

jcastaldoFoodEssentials:
…f writables that the RecordReaderDataSetIterator can support

### DIFF
--- a/datavec-data/datavec-nd4j-common/src/main/java/org/datavec/common/RecordConverter.java
+++ b/datavec-data/datavec-nd4j-common/src/main/java/org/datavec/common/RecordConverter.java
@@ -16,11 +16,11 @@
 
 package org.datavec.common;
 
+import org.datavec.api.writable.DoubleWritable;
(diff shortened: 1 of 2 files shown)


jcastaldoFoodEssentials:
So I played around and took a DataSet and created Writables using thew new function and wrote it to disk using a CSVRecordWriter and I'm finding I can't use a CSVRecordReader with the RecordReaderDataSetIterator to read it back in. 

Here is an example of a line from the saved CSV
[9,28,1,4],3
You'll notice the 3 at the end represents our one hot and [9,28,1,4] is our NDArray. So if you read it in with a comma delimiter, then you get
[9,
28,
1,
4],
3
as separate Writables and we can't parse [9 or 4] with a toDouble call. I'm wondering if I'm saving the data wrong after I generate the Writables or if I'm reading it in wrong.

agibsonccc:
Could you point me at example code for this please?

PR #142, opened 2016-12-19
Use Path instead of File in FileSplit

EdeMeijer:
This is more of a demonstration of how using Path instead of File could improve the code base and make testing easier.


saudet:
But incompatible with Android...

EdeMeijer:
@saudet why is is incompatible? Path is Java 7. Is it not supported on Android?

EdeMeijer:
Okay, I Googled a bit and found that the java.nio.file package is indeed not (fully) supported in Android. That's a bummer. Maybe we can introduce our own little abstraction to make testing easier, but it's probably not worth the hassle.

PR #148, opened 2016-12-24
Transformserver

agibsonccc:
Adds a transform server via rest. This is for integration in to a rest api where you can push raw csvs via rest for scoring.


kepricon:
i think there is still datavec--spark-inference-server information like this
https://github.com/deeplearning4j/DataVec/blob/4fd927320ddfe31c5b272d25f8be1cdacd8dbd8b/pom.xml#L35

I believe we should delete datavec--spark-inference-server module and modify module name in pom.xml(root)

agibsonccc:
This has been fixed. Merged.

PR #161, opened 2017-01-10
Work around conflicting versions of Jackson between GeoIP2 and Spark

saudet:
This doesn't actually work very well. We need a better solution...

### DIFF
--- a/datavec-geo/pom.xml
+++ b/datavec-geo/pom.xml
@@ -27,6 +27,7 @@
 
     <properties>
         <geoip2.version>2.8.0</geoip2.version>
+        <geoip2.jackson.version>2.8.2</geoip2.jackson.version>
     </properties>
 
     <dependencies>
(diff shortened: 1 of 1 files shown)


datavec-geo/pom.xml
```diff
@@ -40,6 +41,11 @@
             <artifactId>geoip2</artifactId>
             <version>${geoip2.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.module</groupId>
+            <artifactId>jackson-module-scala_2.10</artifactId>
+            <version>${geoip2.jackson.version}</version>
```
agibsonccc: @AlexDBlack couldn't we use shading here?

saudet: Not that I know of, there are classes used by GeoIP itself. Unless you mean to shade GeoIP as well? Spark 1.6 seems to work well with newer versions of Jackson anyway, so I'm thinking  we might be able to get away with having users specify that dependency in their pom.xml file to make sure it ends up overriding everything.

PR #171, opened 2017-01-21
Bulding multiple spark major versions 1 and 2

elyast:
Updating scripts to run multiple spark versions

### DIFF
--- a/buildmultiplescalaversions.sh
+++ b/buildmultiplescalaversions.sh
@@ -2,8 +2,10 @@
 set -eu
 ./change-scala-versions.sh 2.10
 mvn "$@"
-mvn -Dspark.version=2.1.0 -Dspark.major.version=2 "$@"
+./change-spark-versions.sh 2
+mvn "$@"
 ./change-scala-versions.sh 2.11
 mvn "$@"
-mvn -Dspark.version=2.1.0 -Dspark.major.version=2 "$@"
+./change-spark-versions.sh 1
+mvn "$@"
 ./change-scala-versions.sh 2.10
--- a/buildmultiplescalaversions.sh
+++ b/buildmultiplescalaversions.sh
@@ -2,6 +2,8 @@
 set -eu
 ./change-scala-versions.sh 2.10
 mvn "$@"
+mvn -Dspark.version=2.1.0 -Dspark.major.version=2 "$@"
 ./change-scala-versions.sh 2.11
 mvn "$@"
+mvn -Dspark.version=2.1.0 -Dspark.major.version=2 "$@"
 ./change-scala-versions.sh 2.10
--- a/buildmultiplescalaversions.sh
+++ b/buildmultiplescalaversions.sh
@@ -2,10 +2,9 @@
 set -eu
 ./change-scala-versions.sh 2.10
 mvn "$@"
-./change-spark-versions.sh 2
-mvn "$@"
 ./change-scala-versions.sh 2.11
 mvn "$@"
-./change-spark-versions.sh 1
+./change-spark-versions.sh 2
 mvn "$@"
 ./change-scala-versions.sh 2.10
+./change-spark-versions.sh 1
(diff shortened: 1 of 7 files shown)


AlexDBlack:
I can't speak to the release script, but the approach looks good to me. I tested buildmultiplescalaversions.sh just to be sure, builds fine.

@agibsonccc @huitseeker - do we want to release Spark 2 with Scala 2.10 at all? Scala 2.10 is deprecated as of Spark 2.1. I'd be happy with just doing 2.11 for Spark 2.


Second issue: the versioning scheme (as of 0.7.3 release anyway) is currently as follows:
groupId / artifactId
org.datavec / datavec-spark_2.1x (or 2.11) for both spark 1 and 2
version:
- 0.7.3_spark_1
- 0.7.3_spark_2

Is this the sort of GAV scheme we want? I would have thought the following might be preferable?
org.datavec / datavec-spark_2.11 / 0.7.3
org.datavec / datavec-spark2_2.11 / 0.7.3

agibsonccc:
We shouldn't need it for spark 2.

elyast:
@AlexDBlack I agree that could be beneficial, however it is common in spark based libraries to suffix spark version in version itself rather than manipulate with artifactId, see spark-testing-base

elyast:
any progress on this ?

perform-release.sh
```diff
@@ -23,8 +23,12 @@ mvn versions:set -DallowSnapshots=true -DgenerateBackupPoms=false -DnewVersion=$
 
 source change-scala-versions.sh 2.10
 mvn clean deploy -Dgpg.executable=gpg2 -DperformRelease -Psonatype-oss-release -DskipTests -DstagingRepositoryId=$STAGING_REPOSITORY
+source change-spark-versions.sh 2
+mvn clean deploy -Dgpg.executable=gpg2 -DperformRelease -Psonatype-oss-release -DskipTests -DstagingRepositoryId=$STAGING_REPOSITORY
```
huitseeker: We won't need a deployment of Spark 2.x.x for Scala 2.10, so that you can move the change of spark version to after the change of scala version and delete the extra deploy.

pom.xml
```diff
@@ -72,8 +72,6 @@
         <freemarker.version>2.3.23</freemarker.version>
         <logback.version>1.1.2</logback.version>
         <maven-deploy-plugin.version>2.8.2</maven-deploy-plugin.version>
-        <spark.version>1.6.2</spark.version>
```
huitseeker: I would have kept the parent pom version stanza rather than the datavec-spark property. That way if we have another project that ends up using spark, there's no duplication.

elyast: I disagree, reason is simple datavec-spark pom will be registered under 2 different versions, with 2 different properties, however datavec-parent pom will have only 1 version registered.
This will backfire later on when trying to use this as dependency

perform-release.sh
elyast: Sure I will remove this

buildmultiplescalaversions.sh
```diff
@@ -4,4 +4,7 @@ set -eu
 mvn "$@"
 ./change-scala-versions.sh 2.11
 mvn "$@"
+./change-spark-versions.sh 2
```
huitseeker: Unfortunately, 71f944405341901feb81f66de265c15160221d82 has changed the default state of the repository. 

So while this script was going through the following state transitions :
(2.10, 1) -> (2.11, 1) -> (2.11, 2) -> (2.10, 1) 
It will now need to implement:
(2.11, 1) -> (2.10, 1) -> (2.11, 2) -> (2.11, 1) or equivalently
(2.11, 1) -> (2.11, 2) -> (2.10, 1) -> (2.11, 1)

saudet: Ah, yes, forgot to update these scripts, sorry about that

elyast: pushed

elyast:
:)

PR #172, opened 2017-01-21
add ndarray conversion

agibsonccc:
Add batch ndarray creation (eg: a matrix) to the inference server. Needs test yet.


agibsonccc:
@skonto @debasishg

PR #191, opened 2017-02-28
Automated Syntax correction and enforcement

huitseeker:
Please refer to https://github.com/deeplearning4j/nd4j/pull/1626 — this is the same idea applied to a different repo.
Enforces https://google.github.io/styleguide/javaguide.html

I'll just repeat how you can get automated formatting in case of an issue:
- run 'mvn formatter:format' in any subproject you want
- use the IDEA plugin https://plugins.jetbrains.com/idea/plugin/8527-google-java-format
- use Eclipse's built-in code formatter with the provided contrib/formatter.xml format file
- if all else fails, replace in the root pom the goal of the mvn-formatter-plugin :
  <goal>validate</goal>
  with
  <goal>format</goal>


huitseeker:
Forgot a file, do not merge

huitseeker:
I can remove the enforcement for now, if need be.

agibsonccc:
Yeah just concerned about how we "onboard" this. We should fix all the enforcement errors in this branch before merging this.Yeah just concerned about how we "onboard" this. We should fix all the enforcement errors in this branch before merging this.

huitseeker:
All the enforcement errors are fixed in 2850231, otherwise I would not have submitted the PR.

huitseeker:
Rather than maintaining this, I'd rather merge the corrections without the enforcement and have a later PR for enforcement.

PR #196, opened 2017-03-03
Move spark version properties into the spark module itself

AlexDBlack:
***WIP DO NOT MERGE***


elyast:
@huitseeker the same needs to be done for deeplearning4j: spark properties needs to be present in spark/dl4j-streaming module rather than root module

AlexDBlack:
@elyast I'm working on the same changes for DL4J
And no worires - I mised this during the earlier reviews. It's a common issue with Maven.

PR #198, opened 2017-03-03
Cleanups and optimizations of Datavec,  ahead of a refactor of transforms & AnalyzeSpark

huitseeker:
- make Quality classes monadic
- remove old `musicg` module
- refactorings, mostly in utils & spark


datavec-api/src/main/java/org/datavec/api/transform/quality/columns/StringQuality.java
```diff
@@ -31,39 +33,57 @@
...
+        /*
+         * The algorithm used is based on streamlib's implementation of "HyperLogLog in Practice:
+         * Algorithmic Engineering of a State of The Art Cardinality Estimation Algorithm", available
+         * <a href="http://dx.doi.org/10.1145/2452376.2452456">here</a>.
+         *
+         * The relative accuracy is approximately `1.054 / sqrt(2^p)`. Setting
+         * a nonzero `sp > p` in HyperLogLogPlus(p, sp) would trigger sparse
+         * representation of registers, which may reduce the memory consumption
+         * and increase accuracy when the cardinality is small.
+         */
```
AlexDBlack: HLL: It's not something I'm familiar with. But overall, using an approximate (but much more scalable) count makes sense.
There are some situations where we need an exact count - usually when the cardinality is small. Though I guess we have functionality for that elsewhere (AnalyzeSpark.getUnique, for example). The main thing I guess is to make it clear to users that it is approximate, not exact.

datavec-spark/pom.xml
```diff
@@ -128,6 +128,12 @@
             <version>${nd4j.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>com.clearspring.analytics</groupId>
+            <artifactId>stream</artifactId>
+            <version>2.7.0</version>
```
AlexDBlack: As per previous version comment.

huitseeker:
@alexdblack the audio I'm deleting here is a duplicate ! Search for Wave.java for example

datavec-api/src/main/java/org/datavec/api/transform/quality/columns/StringQuality.java
agibsonccc: @AlexDBlack TLDR: HLL is standard practice. I'm a big fan of this change.

datavec-api/src/main/java/org/datavec/api/util/ArchiveUtils.java
```diff
@@ -204,6 +124,48 @@ else if (file.endsWith(".gz")) {
...
+                f.mkdirs();
+            }
+            /**
+             * If the entry is a file,write the decompressed file to the disk
+             * and close destination stream.
+             **/
+            else {
+                int count;
+
+                FileOutputStream fos = new FileOutputStream(dest + File.separator + entry.getName());
```
agibsonccc: Maybe an IOUtils.copy here instead?

huitseeker:
Done with PR comments. To reiterate : there are two copies of every audio class I delete in this PR, one in `datavec-data/datavec-data-audio/src/main/java/org/datavec/audio/musicg/`, one in `datavec-data/datavec-data-audio/src/main/java/org/datavec/audio/`. I keep the later.

saudet:
I was thinking of adding new audio classes using FFmpeg, similarly to what I have done with the codec classes. I think we should still keep the original classes around though..

huitseeker:
Guys, I'm happy adding the deletion of audio classes to another PR, it's not like it's complicated to do :)

saudet:
@huitseeker Removing the duplicated classes, I'm fine with that! I was talking about Alex's proposal of removing even more.

PR #200, opened 2017-03-06
Refactor Counters & Quality analysis

huitseeker:
WIP : to undergo performance test. Should only go in master after https://github.com/deeplearning4j/DataVec/pull/198 is merged.


datavec-spark/src/main/java/org/datavec/spark/transform/AnalyzeSpark.java
```diff
@@ -395,21 +160,19 @@ public static DataAnalysis analyze(Schema schema, JavaRDD<List<Writable>> data,
...
-                                    .max(iac.getMaxValueSeen()).mean(((double) iac.getSum()) / iac.getCountTotal())
-                                    .sampleStdev(iac.getSampleStdev()).sampleVariance(iac.getSampleVariance())
-                                    .countZero(iac.getCountZero()).countNegative(iac.getCountNegative())
-                                    .countPositive(iac.getCountPositive()).countMinValue(iac.getCountMinValue())
-                                    .countMaxValue(iac.getCountMaxValue()).countTotal(iac.getCountTotal()).build();
+                                    .max(iac.getMaxValueSeen()).mean(iac.getMean()).sampleStdev(iac.getSampleStdev())
+                                    .sampleVariance(iac.getSampleVariance()).countZero(iac.getCountZero())
+                                    .countNegative(iac.getCountNegative()).countPositive(iac.getCountPositive())
+                                    .countMinValue(iac.getCountMinValue()).countMaxValue(iac.getCountMaxValue())
+                                    .countTotal(iac.getCountTotal()).build();
```
AlexDBlack: It might be cleaner to replace this with a method on IntegerAnalysisCounter, intead of doing this manual counter -> implementation here? Especially now that we are using the StatCounter
Same thing for the other analysis tyes (double, long, etc).

datavec-spark/src/main/java/org/datavec/spark/transform/AnalyzeSpark.java
```diff
@@ -653,13 +376,15 @@ public static DataQualityAnalysis analyzeQuality(final Schema schema, final Java
         data.cache();
         int nColumns = schema.numColumns();
 
-        //This is inefficient, but it's easy to implement. Good enough for now!
+
+        List<ColumnType> columnTypes = schema.getColumnTypes();
+        List<QualityAnalysisState> states = data.aggregate(null, new QualityAnalysisAddFunction(schema),
+                        new QualityAnalysisCombineFunction());
```
AlexDBlack: Not much to add. Obviously much more efficient than the current implementation, the approach used here LGTM.

datavec-spark/src/main/java/org/datavec/spark/transform/SparkTransformExecutor.java
```diff
@@ -295,14 +297,25 @@ public static boolean isTryCatch() {
...
+        for (int i = 0; i < rightColumnNames.length; i++) {
+            rightColumnIndexes[i] = join.getRightSchema().getIndexOfColumn(rightColumnNames[i]);
+        }
+        JavaPairRDD<List<Writable>, List<Writable>> rightJV =
+                        right.mapToPair(new ExtractKeysFunction(rightColumnIndexes));
+
+        JavaPairRDD<List<Writable>, Tuple2<Iterable<List<Writable>>, Iterable<List<Writable>>>> cogroupedJV =
+                        leftJV.cogroup(rightJV);
+
+        return cogroupedJV.flatMap(new ExecuteJoinFromCoGroupFlatMapFunction(join));
```
AlexDBlack: This (and the other join-related changes) LGTM.

saudet:
I'm not familiar with this code so I'll let @AlexDBlack decide.

datavec-spark/src/main/java/org/datavec/spark/transform/AnalyzeSpark.java
huitseeker: IIUC, the `(Integer|String|Long|Double|Categorical)Analysis` class is meant to represent something else than the `(Integer|String|Long|Double|Categorical)AnalysisCounter`, including histograms, and *not* including `add`/`merge`. Note the comment I replace [here](https://github.com/deeplearning4j/DataVec/pull/200/commits/a9e4afd42f9f47dde54a7ae057386ff7ac81e44f#diff-857ce1d778fd640e3234282aca9c55a4R143) : I don't want the difference between what those counters are doing and what `Analysis` is supposed to be to be forgotten.

So the "really clean" solution of retiring `XAnalysis` (¹) isn't quite right IMHO. I can certainly move the call to the `Builder` inside `(Integer|String|Long|Double|Categorical)AnalysisCounter`(²), but that's just passing down the bucket. Another solution would be to have `Analysis` extend `AnalysisCounter` (³) — but `Analysis` is a weird place for the `add`/`merge` logic.

huitseeker:
Closing and reopening with the correct target branch.

PR #211, opened 2017-03-20
Intropro081

smaryn:
Updated POM file with profiles for different repositories.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -59,7 +59,7 @@
         <play.version>2.4.6</play.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <slf4j.version>1.7.7</slf4j.version>
-        <nd4j.version>0.8.1-SNAPSHOT</nd4j.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -59,7 +59,7 @@
         <play.version>2.4.6</play.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <slf4j.version>1.7.7</slf4j.version>
-        <nd4j.version>0.8.1-SNAPSHOT</nd4j.version>
+        <nd4j.version>${project.parent.version}</nd4j.version>
```
saudet: This is the parent pom.xml file. Does that actually work?

huitseeker: Yes: the version is only effectively used in child projects. When they inherit this value of `${project.parent.version}` for `nd4j.version`, it resolves to this pom's (the parent pom's) version. Neat trick.

here's the output of `mvn help:effective-pom`
https://gist.github.com/98e9b288e2d8e0e8705df04a7abd5a6e

Issue #217, opened 2017-03-23
Transforms should be chainable and, optionally, randomly invoked


crockpotveggies:
For the purpose of image classification and learning, transforms like CropImageTransform and WarpImageTransform need to be chainable and have the option of being randomly invoked throughout the chain. This would allow for a more homogenous training set and contribute to a smoother learning curve.

raver119:
We have concept in few nlp places, where we allow combination of abstractions.
Maybe that's what we need here? Some transform holder, that's built from any number of real transforms, and applies them sequentially or with shuffle?

crockpotveggies:
I like the transform holder idea a lot. Shuffle should probably be optional, since some image pipelines may distort an image in a negative way?

crockpotveggies:
@raver119 can you link me to where this is done in NLP?

raver119:
It's just abstraction/pattern, not sure how particular impl is going to help you :)

This: 
https://github.com/deeplearning4j/deeplearning4j/blob/1f8af820c29cc5567a2c5eaa290f094c4d1492a7/deeplearning4j-nn/src/main/java/org/deeplearning4j/datasets/iterator/CombinedPreProcessor.java#L15-L15

Or this: 
https://github.com/deeplearning4j/deeplearning4j/blob/1f8af820c29cc5567a2c5eaa290f094c4d1492a7/deeplearning4j-nlp-parent/deeplearning4j-nlp/src/main/java/org/deeplearning4j/text/sentenceiterator/AggregatingSentenceIterator.java

Idea is the same behind those implementation.
We have class implementing some interface, that holds some other implementations of the same interface, and ivoke them sequentially. If you want to invoke them in random order - it's fairly easy to add  as well.

crockpotveggies:
Pattern helps me to understand what has been done already and trying to stay within our established practices.

I'm thinking we create a `CombinedImageTransform` which can take multiple `ImageTransform` classes. It takes a `List<InvocationEnum, ImageTransform>` which allows the user to define the pipeline and whether that specific element in the pipeline is sequentially or randomly evoked.

Ideas?

raver119:
I'd keep it as List. 

In this case you can pass transform 0 as mandatory for all paths, and transofrm 1 as another combined transform, which will invoke its members in random order

crockpotveggies:
I don't think that'd work though if you, say, always need a random crop but don't always want to randomly warp an image. For example take the following pipeline:

1. (Always) RandomCrop
[...]

PR #218, opened 2017-03-23
Add RandomCrop and CombinedImageTransform.

crockpotveggies:
Allows for random cropping of images and composing image pipelines of random variability.

Addresses https://github.com/deeplearning4j/DataVec/issues/217

Please merge when tests are added.


datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/CombinedImageTransform.java
```diff
@@ -0,0 +1,85 @@
...
+ *  *
+ *  *    Unless required by applicable law or agreed to in writing, software
+ *  *    distributed under the License is distributed on an "AS IS" BASIS,
+ *  *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  *    See the License for the specific language governing permissions and
+ *  *    limitations under the License.
+ */
+package org.datavec.image.transform;
+
+import javafx.util.Pair;
```
raver119: No. No javafx please.

Use our Pair.

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/CombinedImageTransform.java
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @author Justin Long (@crockpotveggies)
+ */
+public class CombinedImageTransform extends BaseImageTransform<opencv_core.Mat> {
+
+    protected List<Pair<ImageTransform, Double>> transforms;
+    protected boolean shuffle;
+
+    public CombinedImageTransform(List<Pair<ImageTransform, Double>> transforms) {
+        this(new Random(1234), transforms, false);
```
raver119: No, not a java Random please.

Use native Nd4j.getRandom() method, and use it instead.

crockpotveggies: there's a problem here. base class for ImageTransform requires a java random. I can still set this and choose not to use it

raver119: Let's update base class then.

Nd4j.getRandom() follows seed/sequence defined within model initialization, so it's kinda nice option to have for reproducibility.

crockpotveggies:
Updated as per Gitter chat. Tests passing.

raver119:
I'll pick it up from here, and add some more methods.

saudet:
Could we also add MultiImageTransform as a specialization of PipelineImageTransform just so we don't break the API?

It sounds like we could also have enhanced CropImageTransform instead of creating another RandomCropTransform class?

raver119:
Damn, yea. I'll do it now.

raver119:
Sorry, it's back now, backed by PupelineImageTransform.

PR #223, opened 2017-03-28
Fix for #3130: Better handling for non existing data in BaseImageRecordReader and/or FileSplit

vlovsky:


vlovsky:
One quirk, I don't know if new File("") is **intended** to be handled as _working dir_ in the FileSplit

AlexDBlack:
Right, my only problem with this: new File("") can be valid - i.e., it's the current/launch directory. Though it might not be best practice, I don't see why a user couldn't have their files in the launch directory?

i.e., this would give an (incorrect) error for a project organized as follows
/some/directory/myJar.jar
/some/directory/imgClass0/
/some/directory/imgClass1/
etc

Maybe we just need a rootDir.exists() check?

vlovsky:
I discarded rootDir.exists(), as I thought that there could be wild card characters in the path. Your solution would be superior if the path _indeed_ must exist as specified without expansion. Please advise.

AlexDBlack:
> there could be wild card characters in the path

FileSplit takes in a single file or a directory (as a File object); FileUtils.listFiles(rootDir, allowFormat, recursive); probably won't work with wildcards, but I haven't tested it. So I don't see wildcards as an issue for now/here, though adding support for them in the future is worth considering.

vlovsky:
Then the best solution for now is to add exists 

if (!rootDir.exists())
  <report and issue>
else if (rootDir.isDirectory()). 

During wild card characters implementation the expansion should probably happen before these two, although the the expansion itself would validate if folders existed. The edge case is when the expansion doesn't translate to anything that exist. And that probably should through an exception.

```
$ ls -lart /c/temp/31*
total 40
drwxr-xr-x 1 vlad vlad    0 Mar 27 10:11 train
drwxr-xr-x 1 vlad vlad    0 Mar 27 10:13 test
drwxr-xr-x 1 vlad vlad    0 Mar 27 10:30 .
-rw-r--r-- 1 vlad vlad 1596 Mar 27 10:35 passengers_raw.csv
drwxr-xr-x 1 vlad vlad    0 Mar 28 22:01 ..

vlad@vlad-pc MINGW64 /c/projects/deeplearning4j
$ ls -lart /c/temp/32*
ls: cannot access '/c/temp/32*': No such file or directory
```

vlovsky:
made a commit for the same...

PR #231, opened 2017-04-13
Remove .transform(null) in BaseImageRecordReader

crockpotveggies:

In hasNext() there were multiple calls to imageTransform.transform(null). From reviewing the code and our transforms, I could not find any benefits to doing so. Removing this code.

## How was this patch tested?

Ran unit tests and runs on my local.

### DIFF
--- a/datavec-data/datavec-data-image/src/main/java/org/datavec/image/recordreader/BaseImageRecordReader.java
+++ b/datavec-data/datavec-data-image/src/main/java/org/datavec/image/recordreader/BaseImageRecordReader.java
@@ -232,19 +232,9 @@ public abstract class BaseImageRecordReader extends BaseRecordReader {
     @Override
     public boolean hasNext() {
         if (iter != null) {
-            boolean hasNext = iter.hasNext();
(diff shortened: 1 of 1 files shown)


crockpotveggies:
I think this was done for ShowImageTransform? But how about we just fix ShowImageTransform and avoid all of these calls to the transformer?

saudet:
If we don't care that the process doesn't terminate when processing is
done, then we can remove that, sure.

saudet:
If we do care, we'll need another mechanism to close the Frame that displays images in ShowImageTransform, for example.

huitseeker:
We could have a NullOpShowImageTransform variant that we'd call on this ?

saudet:
The issue is not the transform itself, but the call to some method of the transform in `RecordReader.hasNext()`. It's supposed to be idempotent method.

One solution I see is forcing users to call some `ImageTransform.close()` method (or try-with-resources)...

saudet:
@AlexDBlack IIRC, you had a similar issue in Arbiter or some such not terminating because of opened windows. What did you end up doing in that case?

PR #233, opened 2017-04-16
Make all image transforms thread-safe.

crockpotveggies:

Because of incoming changes with https://github.com/deeplearning4j/deeplearning4j/pull/3258 all transforms being used by RecordReaderDataSetIterator will need to be thread-safe.

One concern here is will this create a lot of memory pointers? Is that a problem? Are we going to see OOMs with large numbers of images?

## How was this patch tested?

Ran tests and also implemented in long-running training process.

### DIFF
--- a/datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
+++ b/datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
@@ -18,6 +18,9 @@ package org.datavec.image.loader;
 import java.awt.image.BufferedImage;
 import java.io.*;
 import java.nio.ByteOrder;
+import java.util.HashMap;
+import java.util.Map;
+
 import org.apache.commons.io.IOUtils;
 import org.bytedeco.javacpp.DoublePointer;
 import org.bytedeco.javacpp.FloatPointer;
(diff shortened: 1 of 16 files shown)


saudet:
That won't work well. The converters and the display frame weren't meant to be used that way. We could try to use TLS, but why not just use one pipeline per thread?

crockpotveggies:
I like that idea, where do you suggest we tie-in the RRDSI thread with a pipeline? Since ImageRecordReader is thread-unaware, how could we get IRR.next() to recognize which thread it's being called on?

The pattern is currently this:
1. RecordReaderDataSetIterator.next(num) calls ImageRecordReader.next() for specified num of times
2. I've parallelized RRDSI.next(num) by calling each IRR inside a fixed thread pool
3. Essentially the NativeImageLoader inside IRR.next() would need to be aware of the thread pool size.

Otherwise, we need consider a design change of RecordReader?

saudet:
Maybe by adding some sort of "offset" and "skip" parameters where the default is 0 and 1, but where we could specify 1 and 2, etc? We wouldn't need to make thread aware...

crockpotveggies:
Can you explain how that would solve the issue where an issue of OpenCVFrameConverter.toMat would be guaranteed to always see the same instance of the frame being passed to it?

My earlier tests showed there's still risk that the converter can start processing the wrong image frame if you can't guarantee that specific instance is tied to the thread.

The good news is I use a fixed thread pool so we won't be creating more objects than in that thread pool (avoid mem leak). However, do our users access IRR without RRDSI?

saudet:
The RecordReaderDataSetIterator would need to manage the thread pool, obviously, but the RecordReaders would just need to know which files to load and which ones to ignore. It doesn't need to know anything about threads!

saudet:
Sure, users can use RecordReader alone, but it doesn't load more than one image at a time anyway, so that's fine. If we ever need to make loading each image faster, we'll need to optimize the image decoder itself, not the RecordReader.

agibsonccc:
I don't agree with this in general. I'm closing this.

crockpotveggies:
@agibsonccc this approach has already been abandoned in favor of the work completed with @raver119.

PR #234, opened 2017-04-17
Dh 3198 cifar10

kepricon:

in current master, when users try to set bigger size(say, 3x128x128) than it's original size(3x32x32), it has errors like this(https://github.com/deeplearning4j/deeplearning4j/issues/3198)

the cause of this situation was trying to use bigger size(128) to load the original image(3x32x32).

## How was this patch tested?

unit test codes are included in this PR.


kepricon:
Could you(@AlexDBlack @crockpotveggies ) please review this?

kepricon:
@crockpotveggies yup
this is the fix for deeplearning4j/deeplearning4j#3198 (comment)

kepricon:
@crockpotveggies 
for this(https://github.com/deeplearning4j/deeplearning4j/issues/3198#issuecomment-293976785)
I'm not sure the error you discovered was "ArrayIndexOutOfBoundsException",
if so, I've tested varying batch size, it doesn't seem to show the previous errors.

PR #241, opened 2017-04-25
Added largest blob crop transform

antdood:

Made a transform function for cropping the largest blob from an image.

## How was this patch tested?

I ran it with MNIST and Cifar-10 datasets and used ShowImageTransform to validify the crops. Also used drawContours() from the OpenCV API to confirm the right blob was being cropped.

### DIFF
--- a/datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/LargestBlobCropTransform.java
+++ b/datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/LargestBlobCropTransform.java
@@ -1,0 +1,106 @@
+/*-
+ *  * Copyright 2016 Skymind, Inc.
+ *  *
+ *  *    Licensed under the Apache License, Version 2.0 (the "License");
+ *  *    you may not use this file except in compliance with the License.
+ *  *    You may obtain a copy of the License at
+ *  *
+ *  *        http://www.apache.org/licenses/LICENSE-2.0
+ *  *
+ *  *    Unless required by applicable law or agreed to in writing, software
+ *  *    distributed under the License is distributed on an "AS IS" BASIS,
+ *  *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  *    See the License for the specific language governing permissions and
+ *  *    limitations under the License.
+ */
+package org.datavec.image.transform;
+
+import org.bytedeco.javacv.OpenCVFrameConverter;
+import org.datavec.image.data.ImageWritable;
+import org.datavec.image.transform.BaseImageTransform;
+import org.jfree.util.Log;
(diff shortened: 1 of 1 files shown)


datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/LargestBlobCropTransform.java
```diff
@@ -0,0 +1,106 @@
...
+import org.datavec.image.data.ImageWritable;
+import org.datavec.image.transform.BaseImageTransform;
+import org.jfree.util.Log;
+import org.nd4j.linalg.factory.Nd4j;
+import java.util.Random;
+import static org.bytedeco.javacpp.opencv_core.*;
+import static org.bytedeco.javacpp.opencv_imgproc.*;
+
+public class LargestBlobCropTransform extends BaseImageTransform<Mat> {
+
```
agibsonccc: Javadoc explanations would be really helpful here.

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/LargestBlobCropTransform.java
```diff
@@ -0,0 +1,106 @@
...
+    protected org.nd4j.linalg.api.rng.Random rng;
+
+    protected int mode,method,blurWidth,blurHeight,upperThresh,lowerThresh,cannyOrThresh;
+
```
agibsonccc: May not want to abbreviate variable names here (clarity over brevity in this case)

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/LargestBlobCropTransform.java
```diff
@@ -0,0 +1,106 @@
...
+    public LargestBlobCropTransform() {
+        this(null);
+    }
+
+    public LargestBlobCropTransform(Random random) {
+        this(random, CV_RETR_CCOMP, CV_CHAIN_APPROX_SIMPLE, 3, 3, 100, 200 , 0);
+    }
+
+    public LargestBlobCropTransform(Random random, int mode, int method, int blurWidth, int blurHeight, int UpperThresh, int LowerThresh, int CannyOrThresh){
+        super(random);
```
agibsonccc: Wondering if a lombok style builder would be better here (Just add @Builder to the  annotation. This is a lot of variable arguments.

agibsonccc:
Needs a unit test or 2 as well.

agibsonccc:
Any update here? We're almost ready to merge this.

huitseeker:
[Multimedia support for the Transform process](https://trello.com/c/tyLqHQPH/4-multimedia-support-for-the-transform-process)

huitseeker:
Closed in favor of #313

PR #245, opened 2017-05-01
Minor API changes to RecordReader

raver119:
Minor API change to RecordReader:
- batchesSupported() 
- next(num)

This method provides List<Writable> that contains whole batch at once, so DataSet.merge() call isn't needed

### DIFF
--- a/datavec-api/src/main/java/org/datavec/api/records/reader/RecordReader.java
+++ b/datavec-api/src/main/java/org/datavec/api/records/reader/RecordReader.java
@@ -64,11 +64,27 @@ public interface RecordReader extends Closeable, Serializable, Configurable {
     void initialize(Configuration conf, InputSplit split) throws IOException, InterruptedException;
 
     /**
+     * This method returns true, if next(int) signature is supported by this RecordReader implementation.
+     *
+     * @return
+     */
+    boolean batchesSupported();
+
+    /**
+     * This method will be used, if batchesSupported() returns true.
+     *
+     * @param num
+     * @return
+     */
+    List<Writable> next(int num);
+
+    /**
      * Get the next record
      *
      * @return
      */
     List<Writable> next();
+
 
 
     /**
--- a/datavec-data/datavec-data-image/src/main/java/org/datavec/image/recordreader/BaseImageRecordReader.java
+++ b/datavec-data/datavec-data-image/src/main/java/org/datavec/image/recordreader/BaseImageRecordReader.java
@@ -278,7 +278,7 @@ public abstract class BaseImageRecordReader extends BaseRecordReader {
             cnt++;
         }
 
-        INDArray features = Nd4j.createUninitialized(new int[]{cnt, channels, height, width});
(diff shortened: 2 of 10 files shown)


datavec-data/datavec-data-image/src/main/java/org/datavec/image/recordreader/BaseImageRecordReader.java
```diff
@@ -250,6 +254,56 @@ public boolean hasNext() {
...
+        List<Integer> currLabels = new ArrayList<>();
+        while (cnt < num && iter.hasNext()) {
+            File currFile = iter.next();
+            currBatch.add(currFile);
+            if (appendLabel)
+                currLabels.add(labels.indexOf(getLabel(currFile.getPath())));
+            cnt++;
+        }
+
+        INDArray features = Nd4j.createUninitialized(new int[]{cnt, channels, height, width});
```
AlexDBlack: Probably doesn't matter give the default ordering, but maybe this should be a createUninitialized with C order (for strides)?

datavec-api/src/main/java/org/datavec/api/records/reader/RecordReader.java
```diff
@@ -64,13 +64,29 @@
...
+     */
+    boolean batchesSupported();
+
+    /**
+     * This method will be used, if batchesSupported() returns true.
+     *
+     * @param num
+     * @return
+     */
+    List<Writable> next(int num);
```
AlexDBlack: I'm not *totally* happy with this new API, as it's previously always been the case that that a ```List<Writable>``` is a single example. Now it's somewhat opaque to the number of examples in it. Longer term I'd like to review this and maybe replace it with an object called something like ```BatchRecord```.

raver119: Totally agree... I wasn't going to merge this, but then varts bug was found :(
I'm still working on parallel etl, so i'll change it in next PR

datavec-data/datavec-data-image/src/main/java/org/datavec/image/recordreader/BaseImageRecordReader.java
raver119: Agreed.

datavec-api/src/main/java/org/datavec/api/records/reader/RecordReader.java
raver119: I still need the same for CSV stuff anyway.

datavec-dataframe/pom.xml
```diff
@@ -114,6 +114,12 @@
             <artifactId>jfairy</artifactId>
             <version>0.5.2</version>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.google.guava</groupId>
+                    <artifactId>guava</artifactId>
```
benmccann: why exclude guava?

PR #263, opened 2017-05-14
Refactor Reductions / Reducer to make them Aggregable

### LINKED ISSUE #238: Transforms: add multiple-op reductions
(cc @eraly )
Reduction operations: used to combine a set of examples into a single example.
For example, group records by customer and take the average of each column. Or apply a window function to a time series and 

https://github.com/deeplearning4j/DataVec/blob/master/datavec-api/src/main/java/org/datavec/api/transform/reduce/IReducer.java

We have one implementation here:
https://github.com/deeplearning4j/DataVec/blob/master/datavec-api/src/main/java/org/datavec/api/transform/reduce/Reducer.java

However, this reducer implementation has "one reduction per column" functionality only (i.e., input size and output size are both the same).
To add: an IReducer that allows for multiple reductions per column (for example, min, mean, max *and* stdev for every column). This is only possible with Reducer with manual duplication ops first (which isn't really viable for more than a few columns in practice).

The approach I'd probably suggest: leave the current Reducer class as-is, and create something like a MultiOpReducer. The APIs would have to be a bit different between the two (for example, setting multiple reduction ops as the default), though much of the implementation could be shared (MultiOpReduce extends Reduce?)

### LINKED ISSUE #249: Refactor Datavec reducers to make them aggregatable
At the moment, they're Mapreduce-like reducers without map-side combiners. Meaning that to simplify the processing of specifying different operations on different columns, they rely on a large initial `groupByKey` followed by single-machine processing of whatever the aggregate is. That initial `groupByKey` moves around all the data whereas the reducer, by virtue of its being an aggregation, should be able to summarize (shrink) much of it. The solution is to make every reduce operation on every column part of an `aggregateByKey`, where intermediate reductions can occur locally.

Deliverables
Datavec PR w/ tests, possible dl4j-benchmarks runs (hi @kepricon!)
Expected outcomes
much faster TransformProcess execution in DataVec, multiple reductions per column in DataVec #238, insights on how to later achieve #14

huitseeker:

The `groupByKey` + reduction paradigm used in `SparkTransformExecutor` is a killer for any (sp. distributed) performance, requiring disk access and redistribution of all the data before aggregation.

https://databricks.gitbooks.io/databricks-spark-knowledge-base/content/best_practices/prefer_reducebykey_over_groupbykey.html

The current PR proposes a way to combine, both in succession and in parallel, reduce operations on a `Dataset`'s columns. This will allow reduction parallelism, both distributed (Spark) and local (Java 8 streams, someday).

Fixes #249 
Fixes #238 

TODO:
- [x] JavaDoc
- [x] support `ConditionalReduction`s functionality
- [x] support `ignoreInvalidInColmumns` functionality
- [x] integration tests on benchmark data
- [x] retire old `Reducer`
- [x] perhaps remove `util.function` backport machinery and `AggregableFunction` (not really used)
- [x] Variance, PopulationVariance, UncorrectedStdDev (review request) 
~~-  Support and integrate StringReducer~~ Will be part of another PR on Transforms
- [x] Support and integrate SequenceReduce
- [x] support and integrate recent Window reduce

## How was this patch tested?

See attached Unit tests and above TODO.


datavec-api/src/main/java/org/datavec/api/transform/ops/AggregableMultiOp.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.commons.lang3.tuple.Pair;
+import org.datavec.api.writable.ByteWritable;
+import org.datavec.api.writable.Writable;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Created by huitseeker on 5/8/17.
```
agibsonccc: May want some javadoc here.

datavec-api/src/main/java/org/datavec/api/transform/ops/DecoratedBiFunction.java
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Created by huitseeker on 5/1/17.
+ */
+public abstract class DecoratedBiFunction<T, U, R> implements BiFunction<T, U, R> {
+    public abstract R apply(T var1, U var2);
+
+    @Override
+    public <V> BiFunction<T, U, V> andThen(final Function<? super R, ? extends V> var1) {
+        Objects.requireNonNull(var1);
```
agibsonccc: Isn't this java 8 specific?

datavec-api/src/main/java/org/datavec/api/writable/UnsafeWritableInjector.java
```diff
@@ -0,0 +1,27 @@
+package org.datavec.api.writable;
+
+import java.util.function.Supplier;
+
+/**
+ * Created by huitseeker on 5/13/17.
+ */
+public class UnsafeWritableInjector {
+    public static <T> Writable inject (T x){
+        if (x instanceof Integer) {
```
agibsonccc: NDArrayWritable?

datavec-api/src/main/java/org/datavec/api/transform/ops/AggregatorImpls.java
```diff
@@ -0,0 +1,359 @@
...
+
+
+        public void accept(T n) {
+            if (count == 0) {
+                count = 1L;
+                mean = n.doubleValue();
+                variation = 0D;
+            } else {
+                Long newCount = count + 1;
+                Double newMean = mean + (n.doubleValue() - mean) / newCount;
```
agibsonccc: One thing we're missing. In nd4j we have a bias corrected version and a non bias corrected version. We should keep convention consistent across nd4j and datavec for this.
See:
https://github.com/deeplearning4j/nd4j/blob/master/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/accum/Variance.java#L190

huitseeker: What used to occur in this one place is either a call to [`StatUtils.variance`](https://commons.apache.org/proper/commons-math/javadocs/api-3.4/org/apache/commons/math3/stat/StatUtils.html#variance(double[])) or [this](https://github.com/deeplearning4j/DataVec/blob/master/datavec-dataframe/src/main/java/org/datavec/dataframe/reducing/NumericReduceUtils.java#L366) depending on what you call this on. Both return the bias-corrected sample variance (using n - 1 in the denominator), as I do in this PR (I just unfold [Welford's method for computing variance](https://www.johndcook.com/blog/standard_deviation/) to get the aggregability), and which is also what the code you're [linking to](https://github.com/deeplearning4j/nd4j/blob/master/nd4j-backends/nd4j-api-parent/nd4j-api/src/main/java/org/nd4j/linalg/api/ops/impl/accum/Variance.java#L190) does by default. I'll add an option to get the uncorrected variance. `PopulationVariance` should go over well for the uncorrected variance — and map to the right expectations of `StatUtils` users. Little idea of what to call the uncorrected Standard deviation. `UncorrectedStdDev` is fugly. I'll take sugestions.

datavec-api/src/main/java/org/datavec/api/writable/UnsafeWritableInjector.java
huitseeker: Adding dependency on module 'datavec-nd4j-common' will introduce circular dependency between modules 'datavec-api' and 'datavec-nd4j-common'.

I'll throw with a more explicit user message in this case.

datavec-api/src/main/java/org/datavec/api/transform/ops/AggregatorImpls.java
huitseeker: Done in ddb7bd2

turambar:
@huitseeker good use of the word "aggregable." That's all I got for now. ;)

datavec-api/src/main/java/org/datavec/api/transform/reduce/AggregableReductionUtils.java
```diff
@@ -0,0 +1,395 @@
...
+                    res.add(new AggregatorImpls.AggregableSum<Long>());
+                    break;
+                case Stdev:
+                    res.add(new AggregatorImpls.AggregableStdDev<Long>());
+                    break;
+                case UncorrectedStdDev:
+                    res.add(new AggregatorImpls.AggregableUncorrectedStdDev<Long>());
+                    break;
+                case Variance:
+                    res.add(new AggregatorImpls.AggregableVariance<Long>());
```
AlexDBlack: Missing break

datavec-api/src/main/java/org/datavec/api/transform/reduce/MultiOpReducer.java
```diff
@@ -140,72 +135,73 @@ public Schema transform(Schema schema) {
                 ConditionalReduction reduction = conditionalReductions.get(name);
 
                 String outName = reduction.getOutputName();
-                ColumnMetaData m = getMetaForColumn(reduction.getReduction(), name, inMeta);
-                m.setName(outName);
-                newMeta.add(m);
-
+                for (ReduceOp red: reduction.getReductions()){
+                    ColumnMetaData m = getMetaForColumn(red, name, inMeta);
+                    m.setName(outName);
```
AlexDBlack: This doesn't look right to me. You are setting the same name for multiple columns - which is invalid (column names must be unique).

datavec-api/src/main/java/org/datavec/api/transform/reduce/MultiOpReducer.java
```diff
@@ -16,34 +16,30 @@
...
+import org.nd4j.shade.jackson.annotation.JsonProperty;
 
 import java.io.Serializable;
 import java.util.*;
 
 /**
- * A StringReducer is used to take a set of examples and reduce them.
+ * A MultiOpReducer is used to take a set of examples and reduce them.
  * The idea: suppose you have a large number of columns, and you want to combine/reduce the values in each column.<br>
  * StringReducer allows you to specify different reductions for differently for different columns: min, max, sum, mean etc.
```
AlexDBlack: StringReducer

datavec-api/src/main/java/org/datavec/api/transform/reduce/MultiOpReducer.java
```diff
@@ -339,7 +320,22 @@ public Builder keyColumns(String... keyColumns) {
...
+                List<ReduceOp> ops = new ArrayList<>();
+                if (opMap.containsKey(s))
+                    ops.addAll(opMap.get(s));
+                ops.add(op);
+                opMap.put(s, ops);
+            }
+            return this;
+        }
+
+        private Builder addAll(List<ReduceOp> ops, String[] cols) {
```
AlexDBlack: A public version of this (maybe varargs on the String[]) would be useful?

datavec-api/src/main/java/org/datavec/api/transform/reduce/MultiOpReducer.java
```diff
@@ -380,6 +383,25 @@ public Builder stdevColumns(String... columns) {
...
+            return add(ReduceOp.Stdev, columns);
+        }
+        /**
+         * Reduce the specified columns by taking the variance of the values
+         */
+        public Builder variance(String... columns) {
+            return add(ReduceOp.Variance, columns);
+        }
+        /**
+         * Reduce the specified columns by taking the standard deviation of the values
```
AlexDBlack: Javadoc: standard deviation -> variance

datavec-api/src/main/java/org/datavec/api/writable/UnsafeWritableInjector.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Created by huitseeker on 5/13/17.
+ */
+public class UnsafeWritableInjector {
+    public static <T> Writable inject (T x){
+        if (x instanceof INDArray) {
+            throw new IllegalArgumentException("Wrong argument of type INDArray (" +
+                    x.getClass().getName() + ") please use org.datavec.common.data.NDArrayWritable manually to convert.");
+        }
+        else if (x instanceof Integer) {
```
AlexDBlack: As per earlier comment: Integer/Long/etc are final - .getClass() == Integer.class instead?

datavec-api/src/test/java/org/datavec/api/transform/ops/AggregatorImplsTest.java
```diff
@@ -0,0 +1,288 @@
...
+    private List<Integer> intList = new ArrayList<>(Arrays.asList(1, 2, 3, 4, 5, 6, 7, 8, 9));
+    private List<String> stringList = new ArrayList<>(Arrays.asList("arakoa", "abracadabra", "blast", "acceptance"));
+
+    @Test
+    public void AggregableFirstTest(){
+        AggregatorImpls.AggregableFirst<Integer> first = new AggregatorImpls.AggregableFirst<>();
+        for (int i = 0; i < intList.size(); i++){
+            first.accept(intList.get(i));
+        }
+        assertTrue(first.get().toInt() == 1);
```
AlexDBlack: This goes for all of your asserts: use assertEquals(expected, actual) instead.
Main reason: if/when an assert fails, you get a toString (in debugger or test output) with both expected and actual values. With assertTrue, you get no insight into *why* the failure occurred, without manually running the test with a breakpoint.

datavec-api/src/test/java/org/datavec/api/transform/reduce/TestMultiOpReduce.java
```diff
@@ -181,7 +234,7 @@ public void testReduceIntegerIgnoreInvalidValues() {
     }
 
 
-    @Test
+    @Ignore
```
AlexDBlack: Test should be updated, or removed entirely if no longer valid/applicable.

huitseeker:
@AlexDBlack Edited, thanks for the review. Note, e.g. in `testReduceSequenceByWindowFunction` that accumulators that do not encounter a single event now return `NullWritable`.

datavec-api/src/main/java/org/datavec/api/transform/ops/AggregableCheckingOp.java
```diff
@@ -0,0 +1,48 @@
...
+public class AggregableCheckingOp<T> implements IAggregableReduceOp<Writable, T> {
+
+    @Getter
+    private IAggregableReduceOp<Writable, T> operation;
+    @Getter
+    private ColumnMetaData metaData;
+
+    @Override
+    public <W extends IAggregableReduceOp<Writable, T>> void combine(W accu) {
+        if (accu instanceof AggregableCheckingOp) {
```
huitseeker: Exception added.

datavec-api/src/main/java/org/datavec/api/transform/ops/AggregatorImpls.java
```diff
@@ -0,0 +1,423 @@
...
+        private Writable override = null;
+
+        @Override
+        public void accept(T element) {
+            if (element != null) elem = element;
+        }
+
+        @Override
+        public <W extends IAggregableReduceOp<T, Writable>> void combine(W accu) {
+            if (accu instanceof AggregableLast)
```
huitseeker: Exception added.

datavec-api/src/main/java/org/datavec/api/transform/reduce/AggregableReductionUtils.java
huitseeker: added.

datavec-api/src/main/java/org/datavec/api/transform/reduce/MultiOpReducer.java
huitseeker: Done.

agibsonccc:
@AlexDBlack mind a re review here?

huitseeker:
See https://github.com/deeplearning4j/dl4j-examples/pull/460

PR #271, opened 2017-05-17
WIP : Minor fixes for record readers, transforms needed

turambar:

- `TransformProcessRecordReader` calls its member RecordReader's `initialize(...)` method from within its own. This appears to be necessary for using TPRR with, e.g., a CsvRecordReader.
- Added an "ignore unknown tokens" option to `StringListToCategoricalSetTransform`. I'm using this transform to convert a text field into binary BOWs and we want to have the option to ignore tokens that aren't in our vocabulary or replace them with "UNK."

For the latter point, one alternative would be to leave the existing transform unchanged and move this sort of functionality into a dedicated "BOW" transform.

## How was this patch tested?

Manual tests. Shouldn't be hard to add unit tests -- I'll try to get around to that in the coming days.

Please review


turambar:
@emsixteeen you probably want to be working off this branch!

datavec-api/src/main/java/org/datavec/api/transform/transform/string/StringListToCategoricalSetTransform.java
```diff
@@ -119,6 +143,10 @@ public String toString() {
                         + ",categoryTokens=" + categoryTokens + ",delimiter=\"" + delimiter + "\")";
     }
 
+    protected Writable convertToWritable(boolean val) {
```
huitseeker: Is it worth creating a `BooleanWritable` ?

datavec-api/src/main/java/org/datavec/api/transform/transform/string/StringListToCategoricalSetTransform.java
```diff
@@ -138,13 +166,17 @@ public String toString() {
                     String[] split = str.split(delimiter);
                     for (String s : split) {
                         Integer idx = map.get(s);
-                        if (idx == null)
-                            throw new IllegalStateException("Encountered unknown String: \"" + s + "\"");
+                        if (idx == null) {
+                            if (allowUnknownString)
```
huitseeker: Please add a test for this one to [`TestTransforms`](https://github.com/deeplearning4j/DataVec/blob/master/datavec-api/src/test/java/org/datavec/api/transform/transform/TestTransforms.java)

datavec-api/src/main/java/org/datavec/api/records/reader/impl/transform/TransformProcessRecordReader.java
```diff
@@ -38,7 +38,7 @@
      */
     @Override
     public void initialize(InputSplit split) throws IOException, InterruptedException {
-
+        recordReader.initialize(split);
```
huitseeker: You may want to comment on how this step is required here.

turambar:
Most of these have been fixed by other branches so I'm closing it.

PR #274, opened 2017-05-18
Performance improvements for file collections 1M or larger.

crockpotveggies:

Removes unused code in IRR and improves performance for file splits.

## How was this patch tested?

All ImageRecordReader tests passing. This has also been tested in an actual project using 8M image files.


crockpotveggies:
Will merge this in the next day or so if there are no objections.

saudet:
Well, from what I remember, there's a lot of edge cases in that piece of
code. We'll probably end up redebugging a lot of those, which is fine I
guess, but if we don't want future patches to slow down performance
checking for all those weird URLs and what not, we probably want to have an
automated way to test for performance regressions...

PR #282, opened 2017-05-19
Support for zero-based indexing in LibSvm and SVMLight readers

turambar:

As noted [here](http://scikit-learn.org/stable/modules/generated/sklearn.datasets.load_svmlight_file.html), LibSvm and SVMLight files technically allow both zero- and one-based indexing, and both indexing strategies are comon in the wild. Previously, we assumed one-based indexing, which would send us into an infinite loop [here](https://github.com/deeplearning4j/DataVec/blob/master/datavec-api/src/main/java/org/datavec/api/records/reader/impl/misc/LibSvmRecordReader.java#L90).

Now we support both types of indexing, controlled by setting the `ZERO_BASED_INDEXING` property in the reader Configuration.

Note that `ZERO_BASED_INDEXING` is now assumed, so if the file doesn't use it, we will insert an extraneous zero feature, which is OK.

**NOTE: there are a couple places where we use an `assert` or log a warning instead of throwing an informative Exception. This is because adding a new Exception to calls to `next()` requires modifying the RecordReader interface -- I'm passing the buck on that decision. ;)**

## How was this patch tested?

Added a unit test for each type of reader and updated existing unit tests.

Please review


turambar:
Note that `LibSvmRecordReader` and `SVMLightRecordReader` heavily overlap in terms of functionality. I suggest refactoring them to share internals -- just don't have bandwidth to do it.

FWIW, we also don't support multilabel targets, which are allowed in SVMLight, at least.

turambar:
@saudet all righty then, merging.

turambar:
@AlexDBlack crap.

datavec-api/src/main/java/org/datavec/api/records/reader/impl/misc/LibSvmRecordReader.java
```diff
@@ -87,7 +93,13 @@ public void initialize(Configuration conf, InputSplit split) throws IOException,
                 throw new NumberFormatException("Invalid data: " + tokens[k]);
             }
 
-            int j = Integer.valueOf(pair[0]) - 1;
+            int j = Integer.valueOf(pair[0]);
+            if (!zeroBasedIndexing)
+                j = j - 1;
+
+            /* TODO: throw an exception here. */
+            assert(j < 0);
```
huitseeker: Yep, that was not the best idea.

turambar: https://github.com/deeplearning4j/DataVec/pull/286

Issue #134, opened 2016-12-02
Add simple BufferedImageInputSplit


thhart:
I have a couple of Images which I want to process, in my case I split one source into multiple Images and want to process them separately. So all is in memory already. Something like:

`ListBufferedImageInputSplit(Collection<BufferedImage> list)`

The current splits are all very file/uri source related but sometimes the source of the images is not.

This split can be feeded into a ImageRecordReader then.

huitseeker:
![](https://github.trello.services/images/mini-trello-icon.png) [Multimedia support for the Transform process step 1 : images](https://trello.com/c/tyLqHQPH/4-multimedia-support-for-the-transform-process-step-1-images)

raver119:
Is this issue still viable?

saudet:
Still an issue, but I think the DataVec rewrite is going to make this easy to implement @AlexDBlack ?

PR #294, opened 2017-05-25
[WIP] to make transformprocess support images

kepricon:
# do not merge, please

make transformprocess support image data

## How was this patch tested?

unit test codes included


kepricon:
https://github.com/deeplearning4j/DataVec/blob/dh_transformprocess_support_images/datavec-spark-inference-server/src/test/java/org/datavec/spark/transform/ImageSparkTransformTest.java#L22

https://github.com/deeplearning4j/DataVec/blob/dh_transformprocess_support_images/datavec-data/datavec-data-image/src/test/java/org/datavec/image/transform/JsonYamlTest.java#L20

the above tests might be helpful to review this branch.
/cc @huitseeker

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/ResizeImageTransform.java
```diff
@@ -18,6 +18,7 @@
 import org.bytedeco.javacpp.opencv_core;
```
huitseeker: For a transformation this frequent, you may want to expose the choice of interpolation method, see:
http://docs.opencv.org/2.4/modules/imgproc/doc/geometric_transformations.html#void resize(InputArray src, OutputArray dst, Size dsize, double fx, double fy, int interpolation)

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/ScaleImageTransform.java
```diff
@@ -15,23 +15,28 @@
  */
```
huitseeker: This class needs to have maximal sharing with ResizeImageTransform. The Javadoc of each class needs to explain what differences there are between those two classes, and when to use one instead of the other.

datavec-data/datavec-data-image/src/test/java/org/datavec/image/transform/JsonYamlTest.java
```diff
@@ -0,0 +1,97 @@
...
+                .scaleImageTransform(3)
+                .warpImageTransform((float)0.5)
+//              .randomCropTransform(seed, 50, 50)
+//              .filterImageTransform("noise=alls=20:allf=t+u,format=rgba", 100, 100, 4)
+                .build();
+
+        String asJson = itp.toJson();
+        String asYaml = itp.toYaml();
+
+//        System.out.println(asJson);
```
huitseeker: It's better not to leave commented code.

datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
```diff
@@ -539,4 +539,35 @@ protected Mat scalingIfNeed(Mat image, int dstHeight, int dstWidth) {
...
+            if (image == null || image.empty()) {
+                PIX pix = pixReadMem(bytes, bytes.length);
+                if (pix == null) {
+                    throw new IOException("Could not decode image from input stream");
+                }
+                image = convert(pix);
+                pixDestroy(pix);
+            }
+
+            if (converter == null) {
```
huitseeker: This test seems quite redundant. How about initializing the converter in the one constructor that is missing it ? Then you can throw here.

datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
```diff
@@ -539,4 +539,35 @@ protected Mat scalingIfNeed(Mat image, int dstHeight, int dstWidth) {
         }
         return scaled;
     }
+
+    public ImageWritable asWritable(File f) throws IOException {
```
huitseeker: I think the method deserves its own test.

datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
```diff
@@ -539,4 +539,35 @@ protected Mat scalingIfNeed(Mat image, int dstHeight, int dstWidth) {
...
+            if (converter == null) {
+                this.converter = new OpenCVFrameConverter.ToMat();
+//                throw new IOException("need to initialize converter");
+            }
+            ImageWritable writable = new ImageWritable(converter.convert(image));
+            return writable;
+        }
+    }
+
+    public INDArray asMatrix(ImageWritable writable) throws IOException {
```
huitseeker: This needs to explain that it does not rely on `imageTransform` the way `asmatrixView` does, probably with javadoc (or an explicit name, if you can think of it).

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/RotateImageTransform.java
```diff
@@ -59,6 +66,18 @@ public RotateImageTransform(Random random, float angle) {
     /**
      * Constructs an instance of the ImageTransform.
      *
+     * @param centerx maximum deviation in x of center of rotation (relative to image center)
```
huitseeker: Good doc here 👍

datavec-data/datavec-data-image/src/test/java/org/datavec/image/transform/JsonYamlTest.java
```diff
@@ -0,0 +1,97 @@
...
+        ImageTransformProcess itp = new ImageTransformProcess.Builder()
+                .colorConversionTransform(COLOR_BGR2Luv)
+                .cropImageTransform(10)
+                .equalizeHistTransform(CV_BGR2GRAY)
+                .flipImageTransform(0)
+                .resizeImageTransform(300, 300)
+                .rotateImageTransform(30)
+                .scaleImageTransform(3)
+                .warpImageTransform((float)0.5)
+//              .randomCropTransform(seed, 50, 50)
```
huitseeker: What are the issues with those ?

datavec-data/datavec-data-image/src/test/java/org/datavec/image/transform/JsonYamlTest.java
```diff
@@ -0,0 +1,97 @@
...
+        List<ImageTransform> transformList = itp.getTransformList();
+        List<ImageTransform> transformListJson = itpFromJson.getTransformList();
+        List<ImageTransform> transformListYaml = itpFromYaml.getTransformList();
+
+        for (int i = 0; i < transformList.size(); i++) {
+            ImageTransform it = transformList.get(i);
+            ImageTransform itJson = transformListJson.get(i);
+            ImageTransform itYaml = transformListYaml.get(i);
+
+//            System.out.println(i + "\t" + it);
```
huitseeker: Commented prints.

datavec-data/datavec-data-image/src/test/java/org/datavec/image/transform/JsonYamlTest.java
```diff
@@ -0,0 +1,97 @@
...
+//        System.out.println(asYaml);
+
+        ImageWritable img = TestImageTransform.makeRandomImage(0, 0, 3);
+        ImageWritable imgJson = new ImageWritable(img.getFrame().clone());
+        ImageWritable imgYaml = new ImageWritable(img.getFrame().clone());
+        ImageWritable imgAll = new ImageWritable(img.getFrame().clone());
+
+        ImageTransformProcess itpFromJson = ImageTransformProcess.fromJson(asJson);
+        ImageTransformProcess itpFromYaml = ImageTransformProcess.fromYaml(asYaml);
+//
```
huitseeker: Unneeded comment line.

datavec-spark-inference-server/src/main/java/org/datavec/spark/transform/ImageSparkTransform.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+@AllArgsConstructor
+public class ImageSparkTransform {
+    private ImageTransformProcess imageTransformProcess;
+
+    public Base64NDArrayBody toArray(ImageRecord record) throws IOException {
+        ImageWritable record2 = imageTransformProcess.transformFileUriToInput(record.getUri());
+        INDArray finalRecord = imageTransformProcess.executeArray(record2);
+
+//        INDArray convert = RecordConverter.toArray(finalRecord);
```
huitseeker: This is stale w.r.t the `toArray` below, please remove.

datavec-spark-inference-server/src/main/java/org/datavec/spark/transform/ImageSparkTransform.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Created by kepricon on 17. 5. 24.
+ */
+@AllArgsConstructor
+public class ImageSparkTransform {
```
huitseeker: We're going to need a bit more here. Have you thought of having a common interface that ould be implemented by `CSVSparkTransform` and this ? Would you also include an abstract class that would contain just the members shared by this and `CSVSparkTransform` ?

datavec-spark-inference-server/src/main/java/org/datavec/spark/transform/model/BatchImageRecord.java
```diff
@@ -0,0 +1,28 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Created by kepricon on 17. 5. 24.
+ */
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class BatchImageRecord {
```
huitseeker: This will need some JavaDoc.

datavec-spark-inference-server/src/main/java/org/datavec/spark/transform/ImageSparkTransform.java
huitseeker: Finally, some JavaDoc wouldn't go amiss.

datavec-spark-inference-server/src/test/java/org/datavec/spark/transform/ImageSparkTransformTest.java
```diff
@@ -0,0 +1,43 @@
...
+public class ImageSparkTransformTest {
+
+    @Test
+    public void testImageSparkTransform() throws Exception {
+        int seed = 12345;
+
+        File f0 = new ClassPathResource("/testimages/class0/0.jpg").getFile();
+        File f1 = new ClassPathResource("/testimages/class1/A.jpg").getFile();
+
+//        ImageRecord imgRecord = new ImageRecord(f0.toURI());
```
huitseeker: Commented code should be removed.

datavec-spark-inference-server/src/main/java/org/datavec/spark/transform/model/ImageRecord.java
```diff
@@ -0,0 +1,17 @@
...
+
+import java.net.URI;
+
+/**
+ * Created by kepricon on 17. 5. 24.
+ */
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class ImageRecord {
```
huitseeker: So this is just a passthrough for URI for now. While it's not a bad starting point, this will get more interesting as it becomes an interface federating several image formats.

Note https://github.com/deeplearning4j/DataVec/issues/134
Note the list of supported formats here:
https://github.com/deeplearning4j/DataVec/blob/master/datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java#L46

This should give you ideas for image record formats (and subclasses of the interface I mention above).

huitseeker:
[Multimedia support for the Transform process](https://trello.com/c/tyLqHQPH/4-multimedia-support-for-the-transform-process)

huitseeker:
[Multimedia support for the Transform process](https://trello.com/c/tyLqHQPH/4-multimedia-support-for-the-transform-process)

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/ScaleImageTransform.java
kepricon: are you saying 
https://github.com/deeplearning4j/DataVec/blob/dh_transformprocess_support_images/datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/ResizeImageTransform.java#L28
and 
https://github.com/deeplearning4j/DataVec/blob/dh_transformprocess_support_images/datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/ScaleImageTransform.java#L29
 are not sufficient?

datavec-data/datavec-data-image/src/test/java/org/datavec/image/transform/JsonYamlTest.java
kepricon:
```
// Note : since randomCropTransform use random value
                // the results from each case(json, yaml, ImageTransformProcess)
                // can be different
                // don't use the below line
                // if you uncomment it, you will get fail from below assertions
                //  .randomCropTransform(seed, 50, 50)

                // Note : you will get "java.lang.NoClassDefFoundError: Could not initialize class org.bytedeco.javacpp.avutil"
                // it needs to add the below dependency
                // <dependency>
                //     <groupId>org.bytedeco.javacpp-presets</groupId>
                //     <artifactId>ffmpeg-platform</artifactId>
                // </dependency>
                // FFmpeg has license issues, be careful to use it
                //.filterImageTransform("noise=alls=20:allf=t+u,format=rgba", 100, 100, 4)
```
I've added the explanation in the code.

do you think it has to be there? or just remove that comments?

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/ScaleImageTransform.java
huitseeker: yes : 
```
ScaleImageTransform( newWidth - currentWidth, newHeight - currentHeight) == ResizeimageTransform( newWidht, newHeight)
``` 

I have 2 criticisms:
1. while it's worthwile to expose different APIs to what is fundamentally a call to `resize(mat, result, new Size(...))`, if one of the two classes extended the other, the sharing would be more apparent. Nonetheless, it is not the purpose of this PR to refactor code, so it's better to leave this aside for now (perhaps log an issue). 
2. The above invariant should be made explicit through Javadoc

The purpose of these remarks is that no one is puzzled as to which class to use.

datavec-data/datavec-data-image/src/test/java/org/datavec/image/transform/JsonYamlTest.java
huitseeker: That comment is useful & good, thanks !

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/ScaleImageTransform.java
saudet: The purpose of the ScaleImageTransform was to scale by a certain _factor_, not the final size of the image. The idea is then to use something like CropImageTransform to get different parts of the image at different scale. RsizeImageTransform doesn't do anything like that.

kepricon: I added what @saudet  said about @huitseeker's criticisms like below

https://github.com/deeplearning4j/DataVec/pull/294/commits/7dddb592c1f5a78328e013789a298865bdd983ac

datavec-data/datavec-data-image/src/main/java/org/datavec/image/data/ImageWritable.java
```diff
@@ -77,4 +79,21 @@ public long toLong() {
         throw new UnsupportedOperationException();
     }
 
+    @Override
+    public boolean equals(Object obj) {
+        Frame f2 = ((ImageWritable)obj).getFrame();
```
huitseeker: Ouch, I expect a number of `ClassCastException` being thrown here. You should guard this with an `isinstanceof`.

datavec-data/datavec-data-image/src/test/java/org/datavec/image/loader/TestNativeImageLoader.java
```diff
@@ -190,4 +205,30 @@ Mat makeRandomImage(int height, int width, int channels) {
         }
         return img;
     }
+
+    @Test
+    public void testAsWritable() throws Exception {
```
huitseeker: 👍

datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
```diff
@@ -53,7 +49,7 @@
                     "png", "tif", "tiff", "exr", "webp", "BMP", "GIF", "JPG", "JPEG", "JP2", "PBM", "PGM", "PPM", "PNM",
                     "PNG", "TIF", "TIFF", "EXR", "WEBP"};
 
-    OpenCVFrameConverter.ToMat converter = null;
+    protected static OpenCVFrameConverter.ToMat converter = new OpenCVFrameConverter.ToMat();
```
huitseeker: 👍

datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
huitseeker: Nice for the static converter ! Here, you can remove the nullcheck I think — it's even better if you call the converter by its class.

datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
huitseeker: This one is still standing I'm afraid.

datavec-data/datavec-data-image/src/test/java/org/datavec/image/transform/JsonYamlTest.java
```diff
@@ -0,0 +1,110 @@
...
+            ImageTransform itJson = transformListJson.get(i);
+            ImageTransform itYaml = transformListYaml.get(i);
+
+            System.out.println(i + "\t" + it);
+
+            img = it.transform(img);
+            imgJson = itJson.transform(imgJson);
+            imgYaml = itYaml.transform(imgYaml);
+
+            if (it instanceof RandomCropTransform) {
```
huitseeker: 👍

datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
kepricon: sorry I missed this comment

I added javadoc for it. 
I'm not sure you want me to add some comments about the reason why it doesn't rely on `imageTransform`.
If it is your intention, like other methods such as asMatrix(...), asRowVector(...), I didn't mention whether using `ImageTransform` or not.

if you think it still needs to explain the reason or if I misunderstand your meaning, please let me know. :)

datavec-data/datavec-data-image/src/main/java/org/datavec/image/data/ImageWritable.java
```diff
@@ -77,4 +79,25 @@ public long toLong() {
...
+            Frame f2 = ((ImageWritable) obj).getFrame();
+
+            Buffer[] b1 = this.frame.image;
+            Buffer[] b2 = f2.image;
+
+            if (b1.length != b2.length)
+                return false;
+
+            for (int i = 0; i < b1.length; i++) {
+                if (b1[i].equals(b2[i]) == false)
```
crockpotveggies: Nitpick: you could use shorthand and make this a little more elegant. `if (!b1[i].equals(b2[i]))`

datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
crockpotveggies: If you make this static, what if I'm using NativeImageLoader on multiple threads? What was the motivation for this?

datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
```diff
@@ -110,7 +106,6 @@ public NativeImageLoader(int height, int width, int channels, boolean centerCrop
     public NativeImageLoader(int height, int width, int channels, ImageTransform imageTransform) {
         this(height, width, channels);
         this.imageTransform = imageTransform;
-        this.converter = new OpenCVFrameConverter.ToMat();
```
crockpotveggies: Again see previous issue about having multiple NIL on different threads.

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/CropImageTransform.java
```diff
@@ -15,21 +15,28 @@
  */
```
crockpotveggies: I don't see a javadoc for the constructors? Did I miss it?

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/FilterImageTransform.java
```diff
@@ -30,9 +33,34 @@
  * @author saudet
  * @see FFmpegFrameFilter
  */
+@JsonIgnoreProperties({"filter", "converter"})
+@JsonInclude(JsonInclude.Include.NON_NULL)
 public class FilterImageTransform extends BaseImageTransform {
 
-    FFmpegFrameFilter filter;
+    /**
+     * see the javadoc comments of this class
```
crockpotveggies: Style nitpick: I think this would be cleaner to document these fields for the all-args constructor, like so:
```
/**
 * @param filters filter strings to use.
....
```

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/ImageTransform.java
```diff
@@ -15,15 +15,32 @@
...
+        @JsonSubTypes.Type(value = CropImageTransform.class, name = "CropImageTransform"),
+        @JsonSubTypes.Type(value = EqualizeHistTransform.class, name = "EqualizeHistTransform"),
+        @JsonSubTypes.Type(value = FilterImageTransform.class, name = "FilterImageTransform"),
+        @JsonSubTypes.Type(value = FlipImageTransform.class, name = "FlipImageTransform"),
+        @JsonSubTypes.Type(value = RandomCropTransform.class, name = "RandomCropTransform"),
+        @JsonSubTypes.Type(value = ResizeImageTransform.class, name = "ResizeImageTransform"),
+        @JsonSubTypes.Type(value = RotateImageTransform.class, name = "RotateImageTransform"),
+        @JsonSubTypes.Type(value = ScaleImageTransform.class, name = "ScaleImageTransform"),
+        @JsonSubTypes.Type(value = WarpImageTransform.class, name = "WarpImageTransform")})
+
```
crockpotveggies: Just wondering, what specific problem required adding Json annotations for ImageTransform? Are you serializing it?

datavec-data/datavec-data-image/src/main/java/org/datavec/image/data/ImageWritable.java
kepricon: it was the coding style from my previous company.
I will change it and will use shorthand from now on :)

thanks

datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
kepricon: I thought that `OpenCVFrameConverter.ToMat()` was just static converter which can convert something to `Mat`. I had not sufficient time to think about it thoroughly.

I think I need to make this.converter as a non-static field since OpenCVFrameConverter.ToMat() has member fields in its parent class. 
Do you think it is able to resolve your worries?

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/CropImageTransform.java
kepricon: I added javadocs for each member field like below.
https://github.com/deeplearning4j/DataVec/blob/dh_transformprocess_support_images/datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/CropImageTransform.java#L52

and all constructors have simple javadocs about which method will be used.

Do you think I should put more information on it?

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/ImageTransform.java
kepricon: yup we need to serialize it for SKIL.
https://github.com/deeplearning4j/DataVec/blob/dh_transformprocess_support_images/datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/ImageTransformProcess.java#L113

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/FilterImageTransform.java
kepricon: are you saying removing the current javadocs for these fields and adding to the all-args constructors's javadoc instead?

datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
saudet: Yeah, you don't want to make this static.

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/FilterImageTransform.java
crockpotveggies: Yes.

datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
crockpotveggies: The worry is that if there are multiple instances on different threads, then that toMat() class will be reused and hence will throw errors due to corrupt data. So whatever solves that, solves my worries :)

kepricon:
@crockpotveggies I really appreciate the review
like you said, I added thread-safe features fd25650
could you give a comment about this?

saudet:
@crockpotveggies @kepricon Do we really want or need to use thread-local storage? What's the use case?

saudet:
@crockpotveggies Seriously? Could you explain why you need thread-local storage?

agibsonccc:
Yeah thread local storage bothers me a bit. It's error prone and a huge source of memory leaks in java..

crockpotveggies:
One of the previous issues I encountered was that toMat() is not thread safe, and significant errors will occur if there's more than one instance of NativeImageLoader being used. I asked DH to change the converter field from static to non-static, but if there's a more elaborate ETL pipeline set up, the thinking there was to allow NIL be used without issues. Now, if we know that NIL instances will be thread-bound, then yea I agree with @agibsonccc @saudet

agibsonccc:
Could we get this merged some time soon?

huitseeker:
@kepricon 👍

Issue #235, opened 2017-04-18
reading standard csv leads to `no more elements found!`


abossenbroek:
#### Issue Description

I am trying to load a simple csv with five columns and remove one column. When I try this I get a `java.util.NoSuchElementException: No more elements found!`

The code is as follows:
```java
import org.apache.spark.SparkConf;
import org.apache.spark.api.java.JavaRDD;
import org.apache.spark.api.java.JavaSparkContext;
import org.datavec.api.records.reader.RecordReader;
import org.datavec.api.records.reader.impl.csv.CSVRecordReader;
import org.datavec.api.transform.TransformProcess;
import org.datavec.api.transform.schema.Schema;
import org.datavec.api.util.ClassPathResource;
import org.datavec.api.writable.Writable;
import org.datavec.spark.transform.SparkTransformExecutor;
import org.datavec.spark.transform.misc.StringToWritablesFunction;
import org.datavec.spark.transform.misc.WritablesToStringFunction;


import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

import java.util.Arrays;
import java.util.List;

public class AutoEncoder {

    private static Logger log = LoggerFactory.getLogger(AutoEncoder.class);

    public static void main(String[] args) throws Exception {
        int seed = 49;
        int batchSize = 200;
        int iterations = 1;

        log.info("Load data....");
        // First: get the dataset using the record reader. CSVRecordReader handles loading/parsing
        int numLinesToSkip = 1;
        String delimiter = ",";

        Schema schema = new Schema.Builder()
                .addColumnLong("hash")
                .addColumnLong("code")
                .addColumnDouble("costs")
                .addColumnDouble("days")
                .addColumnCategorical("weekday", Arrays.asList("Mo", "Tu", "We", "Th", "Fr", "Sa", "Su"))
                .build();

       //Print out the schema:
        System.out.println("Input data schema details:");
        System.out.println(schema);

        System.out.println("\n\nOther information obtainable from schema:");
        System.out.println("Number of columns: " + schema.numColumns());
        System.out.println("Column names: " + schema.getColumnNames());
        System.out.println("Column types: " + schema.getColumnTypes());


        TransformProcess tp = new TransformProcess.Builder(schema)
                .removeColumns("costs")
                .build();

        //After executing all of these operations, we have a new and different schema:
        Schema outputSchema = tp.getFinalSchema();

        System.out.println("\n\n\nSchema after transforming data:");
        System.out.println(outputSchema);


[...]

PR #300, opened 2017-05-31
CSVRecordReader now overrides hasNext()

turambar:

CSVRecordReader needs to override hasNext() rather than inherit it from LineRecordReader. This is because CSVRecordReader can skip >0 lines, while LineRecordReader cannot. Thus, for CSVs with N lines where we skip N lines, e.g., a CSV with only a header, hasNext() will return true (LineRecordReader sees the header), but next() will throw an exception because CSVRecordReader skips it and then cannot find any non-header lines.

## How was this patch tested?

All CSVRecordReader unit tests pass, and I added an additional test to make sure that hasNext() returns false when we skip all lines.

Please review

### DIFF
--- a/datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVRecordReader.java
+++ b/datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVRecordReader.java
@@ -95,8 +95,7 @@ public class CSVRecordReader extends LineRecordReader {
         this.quote = conf.get(QUOTE, null);
     }
 
-    @Override
-    public boolean hasNext() {
+    private boolean skipLines() {
         if (!skippedLines && skipNumLines > 0) {
             for (int i = 0; i < skipNumLines; i++) {
                 if (!super.hasNext()) {
@@ -106,20 +105,18 @@ public class CSVRecordReader extends LineRecordReader {
             }
             skippedLines = true;
         }
-        return super.hasNext();
+        return true;
+    }
+
+    @Override
+    public boolean hasNext() {
+        return skipLines() && super.hasNext();
     }
 
     @Override
     public List<Writable> next() {
-        if (!skippedLines && skipNumLines > 0) {
-            for (int i = 0; i < skipNumLines; i++) {
-                if (!hasNext()) {
-                    return new ArrayList<>();
-                }
-                super.next();
-            }
-            skippedLines = true;
-        }
+        if (!skipLines())
+            return new ArrayList<>();
         Text t = (Text) super.next().iterator().next();
         String val = t.toString();
         return parseLine(val);
--- a/datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVRecordReader.java
+++ b/datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVRecordReader.java
@@ -18,20 +18,17 @@ package org.datavec.api.records.reader.impl.csv;
 
 
 
-import org.apache.commons.io.IOUtils;
 import org.datavec.api.conf.Configuration;
 import org.datavec.api.records.Record;
 import org.datavec.api.records.metadata.RecordMetaData;
 import org.datavec.api.records.metadata.RecordMetaDataLine;
 import org.datavec.api.records.reader.impl.LineRecordReader;
-import org.datavec.api.split.StringSplit;
 import org.datavec.api.writable.Text;
 import org.datavec.api.split.InputSplit;
 import org.datavec.api.writable.Writable;
 
 import java.io.DataInputStream;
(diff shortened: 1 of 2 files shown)


turambar:
relevant to https://github.com/deeplearning4j/DataVec/issues/235

datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVRecordReader.java
```diff
@@ -93,6 +96,20 @@ public void initialize(Configuration conf, InputSplit split) throws IOException,
     }
 
     @Override
+    public boolean hasNext() {
+        if (!skippedLines && skipNumLines > 0) {
+            for (int i = 0; i < skipNumLines; i++) {
+                if (!super.hasNext()) {
+                    return false;
```
huitseeker: While you're here, can you make it throw `NoSuchElementException` in the same condition, but in `next()` ?

datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVRecordReader.java
```diff
@@ -18,17 +18,20 @@
 
 
 
+import org.apache.commons.io.IOUtils;
```
huitseeker: Unneeded import.

datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVRecordReader.java
```diff
@@ -18,17 +18,20 @@
...
 import org.datavec.api.conf.Configuration;
 import org.datavec.api.records.Record;
 import org.datavec.api.records.metadata.RecordMetaData;
 import org.datavec.api.records.metadata.RecordMetaDataLine;
 import org.datavec.api.records.reader.impl.LineRecordReader;
+import org.datavec.api.split.StringSplit;
```
huitseeker: Unneeded import.

datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVRecordReader.java
```diff
@@ -18,17 +18,20 @@
...
 import org.datavec.api.writable.Text;
 import org.datavec.api.split.InputSplit;
 import org.datavec.api.writable.Writable;
 
 import java.io.DataInputStream;
 import java.io.IOException;
+import java.io.InputStreamReader;
```
huitseeker: Unneeded import.

datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVRecordReader.java
```diff
@@ -93,6 +96,20 @@ public void initialize(Configuration conf, InputSplit split) throws IOException,
...
     @Override
+    public boolean hasNext() {
+        if (!skippedLines && skipNumLines > 0) {
+            for (int i = 0; i < skipNumLines; i++) {
+                if (!super.hasNext()) {
+                    return false;
+                }
+                super.next();
+            }
+            skippedLines = true;
```
huitseeker: I would put this in a `finally` block of which the above loop would be the `try`. It defends against:
- `super.next()` throwing
- early returns of `false` in the case where `!super.hasNext()` not setting `skippedLines`.

turambar:
@huitseeker removed the unneeded imports. Also made some fixes: the "skip lines" functionality is now isolated in a helper function that is called by both hasNext() and next(), and the functions are now consistent in their behavior (next() throws an exception when hasNext() is false).

saudet:
Looks good, but as @huitseeker mentions hasNext() should probably return false instead of throwing an exception though.

datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVRecordReader.java
```diff
@@ -92,17 +92,28 @@ public void initialize(Configuration conf, InputSplit split) throws IOException,
...
+
+    @Override
+    public boolean hasNext() {
+        return skipLines() && super.hasNext();
+    }
+
+    @Override
+    public List<Writable> next() {
+        if (!skipLines())
+            return new ArrayList<>();
```
huitseeker: This is an empty iterator and is expected to throw `NoSuchElementException`.

turambar:
@saudet fixed: https://github.com/deeplearning4j/DataVec/pull/300/files#diff-f0aab34172f52705c1bc9883c6dd9b94R110

@huitseeker fixed: https://github.com/deeplearning4j/DataVec/pull/300/files#diff-f0aab34172f52705c1bc9883c6dd9b94R116

Issue #289, opened 2017-05-24
Add Schema and Transform support for NDArrayWritable


turambar:
#### Issue Description

We currently lack an efficient column-based String Transform for converting a text field to bag-of-words (BOWs). We do have [StringListToCategoricalSet](https://github.com/deeplearning4j/DataVec/blob/master/datavec-api/src/main/java/org/datavec/api/transform/transform/string/StringListToCategoricalSetTransform.java), but it's not scalable to large vocabularies.

I discussed this with @AlexDBlack, and we decided the principled approach is to add NDArrayWritable as a "type" for columns. This requires support in both Schema and Transform. It will also have implications for analysis and saving to disk. When saving to, e.g., a CSV, we think the right approach is to encode the NDArray object using Base64 and then write the Base64 string into the CSV. That way when the CSV is read from disk, we can properly decode it back into an NDArray.

Once this is added, we can, for example, create a StringToBOWTransform that wraps the BOW Vectorizer, etc.

#### Version Information

All versions

#### Contributing

If you'd like to help us fix the issue by contributing some code, but would
like guidance or help in doing so, please mention it!

turambar:
As discussed, @AlexDBlack

AlexDBlack:
Some initial thoughts
- Restrict NDArrays to the same size, and specify this size in the schema
- AnalyzeSpark etc - probably just ignore NDArrayWritable columns
- Threading: we'll be passing arrays between threads constantly (with something like spark). Need to make sure we don't run into any issues as a result.

Some alternative approaches:
- java arrays (int[], float[] etc) Writable objects
- a general-purpose array type (could be useful and much more efficient, for example, for bitsets and the like)
- record binarization (i.e., binary storage format) https://github.com/deeplearning4j/DataVec/issues/77

AlexDBlack:
Implemented here: https://github.com/deeplearning4j/DataVec/pull/302

PR #302, opened 2017-06-02
Transforms: NDArrayWritable support

AlexDBlack:
***WIP DO NOT MERGE*** (Other than reviews, should be pretty much good to go)

Addresses: https://github.com/deeplearning4j/DataVec/issues/289

- Adds NDArray support for Transforms: column type, schema, analysis, analysis HTML export etc
- Adds NDArray math ops:
    - NDArrayScalarOpTransform: scalar A.op(x), where A in an NDArray, x is a scalar, and op is element-wise add/subtract/multiply etc
    - NDArrayColumnsMathOpTransform: essentially does A.op(B), where A and B are INDArrays and op is element-wise add/substract/multiply etc
    - NDArrayMathFunctionTransform: for element-wise sin/cos/abs/log etc on NDArrayWritable
    - NDArrayDistanceTransform: cosine/euclidean/manhattan distances between two NDArray columns
- Make NDArrayWritable support multi-dimensional arrays (not just row vectors). This resulted in some fairly extensive changes to NDArrayWritable hashCode, compareTo, equals methods
- Moved NDArrayWritable from datavec-nd4j-common to datavec-api; completely removed datavec-nd4j-common module as a result
- Tested:
    - New NDArray transforms, schema, TransformProcess etc
    - SparkStorageUtils (map/sequencefile) + NDArrayWritable
    - Analysis w/ NDArrayWritable, HTML analysis export
    - To/from JSON with NDArray columns/transforms

Known issue:
- Clearly, things like WritablesToString function won't work with NDArray columns. No checks for NDArrayWritable objects have been implemented in this (and related) methods; I'm recommending using MapFile/SequenceFile instead of CSV (via SparkStorageUtils, which has been tested with NDArrayWritables)


datavec-data/datavec-nd4j-common/src/main/java/org/datavec/api/transform/ndarray/NDArrayColumnsMathOpTransform.java
```diff
@@ -0,0 +1,129 @@
...
+                        + Arrays.toString(meta.getShape()) + " and " + Arrays.toString(meta2.getShape()));
+            }
+        }
+
+        return new NDArrayMetaData(newColumnName, meta.getShape());
+    }
+
+    @Override
+    protected Writable doOp(Writable... input) {
+        INDArray out = ((NDArrayWritable) input[0]).get().dup();
```
agibsonccc: So this conflicts what I'm seeing above. We seem to already depend on nd4j here. Is there something I'm missing with reflection?

datavec-data/datavec-nd4j-common/src/main/java/org/datavec/api/transform/ndarray/NDArrayColumnsMathOpTransform.java
```diff
@@ -0,0 +1,129 @@
...
+
+        switch (mathOp) {
```
agibsonccc: What about similarity/distances?

datavec-data/datavec-nd4j-common/src/main/java/org/datavec/api/transform/ndarray/NDArrayColumnsMathOpTransform.java
```diff
@@ -0,0 +1,129 @@
...
+            case ScalarMin:
+            case ScalarMax:
+                throw new IllegalArgumentException(
+                        "Invalid MathOp: cannot use " + mathOp + " with NDArrayColumnsMathOpTransform");
+            default:
+                throw new RuntimeException("Unknown MathOp: " + mathOp);
+        }
+
+        //To avoid threading issues...
+        Nd4j.getExecutioner().commit();
```
agibsonccc: Workspaces might be a factor here?

datavec-data/datavec-nd4j-common/src/main/java/org/datavec/api/transform/ndarray/NDArrayMathFunctionTransform.java
```diff
@@ -0,0 +1,148 @@
...
+ * Operations are performed element-wise on each value in the INDArray; operations are specified by {@link MathFunction}
+ *
+ * @author Alex Black
+ */
+public class NDArrayMathFunctionTransform extends BaseColumnTransform {
+
+    //Can't guarantee that the writable won't be re-used, for example in different Spark ops on the same RDD
+    private static final boolean DUP = true;
+
+    private final MathFunction mathFunction;
```
agibsonccc: Maybe we should have categories for "math functions"? Eg: reduce,transform,scalar,..

datavec-data/datavec-nd4j-common/src/main/java/org/datavec/api/transform/ndarray/NDArrayMathOpTransform.java
```diff
@@ -0,0 +1,126 @@
...
+import org.nd4j.shade.jackson.annotation.JsonProperty;
+
+/**
+ * Perform an NDArray/scalar element wise operation, such as X.addi(scalar).
+ * Element wise operations are performed in place on each value of the underlying INDArray
+ *
+ * @author Alex Black
+ */
+public class NDArrayMathOpTransform extends BaseColumnTransform {
+
```
agibsonccc: This feels like it could be called scalar.

datavec-data/datavec-nd4j-common/src/main/java/org/datavec/common/data/NDArrayWritable.java
```diff
@@ -135,100 +115,100 @@ public boolean equals(Object o) {
...
-            } else if (type1 == DataBuffer.Type.FLOAT.ordinal()) {
-                return buffer1.asFloatBuffer().compareTo(buffer2.asFloatBuffer());
-            } else if (type1 == DataBuffer.Type.INT.ordinal()) {
-                return buffer1.asIntBuffer().compareTo(buffer2.asIntBuffer());
-            } else {
-                throw new UnsupportedOperationException("Unsupported data type: " + type1);
+        }
+
+        //At this point: same rank, length, shape
+        NdIndexIterator iter = new NdIndexIterator('c', array.shape());
```
agibsonccc: why not just a direct array.equals comparison?

datavec-data/datavec-nd4j-common/src/main/java/org/datavec/api/transform/ndarray/NDArrayColumnsMathOpTransform.java
AlexDBlack: nd4j-api is now (recently) a dependency of datavec-api - but NDArrayWritable is defined in datavec-nd4j-common. The transforms are defined in datavec-nd4j-common as a result, but the metadata etc is defined in core.
2 options:
(a) move everything - including the NDArrayWritable and new transforms - to datavec-api
(b) Use reflection and keep everything as it currently is
I don't really mind either way.

datavec-data/datavec-nd4j-common/src/main/java/org/datavec/api/transform/ndarray/NDArrayColumnsMathOpTransform.java
AlexDBlack: Sure, why not.

datavec-data/datavec-nd4j-common/src/main/java/org/datavec/api/transform/ndarray/NDArrayColumnsMathOpTransform.java
AlexDBlack: As far as threading - should be OK. As far as workspaces: I don't think we'll be executing these within a workspace (but if we ever do: definitely need some changes)

datavec-data/datavec-nd4j-common/src/main/java/org/datavec/api/transform/ndarray/NDArrayMathFunctionTransform.java
AlexDBlack: Maybe. The MathFunction enum is just sin/sqrt/log etc.
https://github.com/deeplearning4j/DataVec/blob/master/datavec-api/src/main/java/org/datavec/api/transform/MathFunction.java

datavec-data/datavec-nd4j-common/src/main/java/org/datavec/api/transform/ndarray/NDArrayMathOpTransform.java
AlexDBlack: Yup, I'll change that. I was following the existing datavec transform name conventions, but ND4J convention for this (or, just calling it scalar in general) is better.

datavec-data/datavec-nd4j-common/src/main/java/org/datavec/common/data/NDArrayWritable.java
AlexDBlack: It's in the ```int compareTo(@Nonnull Object o)``` method - i.e., for sorting if NDArray is used as a key. Won't be used much in practice but is needed for Writable interface jsut in case.

PR #313, opened 2017-06-09
Follow-up on Adding largest blob crop transform 

kepricon:

add LargestBlobCropTransform with applying @agibsonccc 's comments to DataVec
(the original pr is https://github.com/deeplearning4j/DataVec/pull/241)

since @antdood has no response, @huitseeker and I decided to continue it on a DataVec repo.

@agibsonccc 
for this comment https://github.com/deeplearning4j/DataVec/pull/241#discussion_r113336654
I basically agree with you `this is a lot of variable arguments`, 
but how about leaving it without a Lombok style builder in order to keep consistency with other ImageTransforms?
Do you still think it needs to be applied builder style? 

## How was this patch tested?

test codes are included in `TestImageTransform`


datavec-data/datavec-data-image/src/test/java/org/datavec/image/transform/TestImageTransform.java
```diff
@@ -257,6 +254,25 @@ public void testProbabilisticPipelineTransform() throws Exception {
...
+        showOrig.transform(writable);
+
+        ImageTransform transform = new LargestBlobCropTransform();
+        ImageWritable w = transform.transform(writable);
+
+        ImageTransform showTrans = new ShowImageTransform("Largest Blob", 50);
+        showTrans.transform(w);
+        Frame newFrame = w.getFrame();
+
+        assertNotEquals(frame, newFrame);
```
agibsonccc: In the test itself I think we could improve the assertions here.  Is there any way we could verify what the biggest blob is in advance?

kepricon: as we discussed, i modified the test code with a specific image.

kepricon:
since this test code use updated `NativeImageLoader`'s features from https://github.com/deeplearning4j/DataVec/pull/294

NOTE:
This PR should be merged after https://github.com/deeplearning4j/DataVec/pull/294

If the review passed, I will merge after 294.
/cc @agibsonccc  @huitseeker

Issue #326, opened 2017-06-20
NotSerializableException for DispatchOp when using reduce in TransformProcess


turambar:
#### Issue Description

Using a reduce operation inside of a TransformProcess throws a `java.io.NotSerializableException` for `org.datavec.api.transform.ops.DispatchOp`.

To reproduce, run the DataVec [PivotExample.java](https://github.com/deeplearning4j/dl4j-examples/blob/master/datavec-examples/src/main/java/org/datavec/transform/basic/PivotExample.java) using the latest 0.8.1-SNAPSHOT from the snapshots repository. You'll get this error:

```
Exception in thread "main" java.io.NotSerializableException: org.datavec.api.transform.ops.DispatchOp
Serialization stack:
	- object not serializable (class: org.datavec.api.transform.ops.DispatchOp, value: org.datavec.api.transform.ops.DispatchOp@68dd39d2)
	at org.apache.spark.serializer.SerializationDebugger$.improveException(SerializationDebugger.scala:40)
	at org.apache.spark.serializer.JavaSerializationStream.writeObject(JavaSerializer.scala:47)
	at org.apache.spark.serializer.JavaSerializerInstance.serialize(JavaSerializer.scala:101)
	at org.apache.spark.rdd.PairRDDFunctions$$anonfun$aggregateByKey$1.apply(PairRDDFunctions.scala:168)
	at org.apache.spark.rdd.PairRDDFunctions$$anonfun$aggregateByKey$1.apply(PairRDDFunctions.scala:166)
	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:150)
	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:111)
	at org.apache.spark.rdd.RDD.withScope(RDD.scala:316)
	at org.apache.spark.rdd.PairRDDFunctions.aggregateByKey(PairRDDFunctions.scala:166)
	at org.apache.spark.rdd.PairRDDFunctions$$anonfun$aggregateByKey$3.apply(PairRDDFunctions.scala:206)
	at org.apache.spark.rdd.PairRDDFunctions$$anonfun$aggregateByKey$3.apply(PairRDDFunctions.scala:206)
	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:150)
	at org.apache.spark.rdd.RDDOperationScope$.withScope(RDDOperationScope.scala:111)
[...]

PR #328, opened 2017-06-20
WIP Mark IAggregableReduceOp as Serializable.

wmeddie:

Mark IAggregableReduceOp as Serializable to fix running Reducers on Spark.

## How was this patch tested?

Added a test that uses reflection to check that every ...Op or Aggregable...


wmeddie:
Fixes #326

wmeddie:
@agibsonccc Is this test OK? It might be a little fragile.  Open to ideas.

PR #336, opened 2017-06-27
DCK poc transforms

turambar:
**NOTE: Replaced by https://github.com/deeplearning4j/DataVec/pull/340.**

Adding some transforms:
- ChangeCaseStringTransform: change case of string column
- ConcatenateStringColumns: concatenate multiple string columns into one
- StringListToCountsNDArrayTransform: string column to dense BOW NDArray
- StringListToIndicesNDArrayTransform: string column to sparse BOW NDArray (i.e., list of indices)

and one Spark function:
- WritablesToNDArrayFunction: convert a list of writables to an NDArray

## How was this patch tested?

Added unit tests ChangeCaseStringTransform and ConcatenateStringColumns to TestTransforms unit test class. Haven't added unit tests for StringList... transforms.

Please review


datavec-api/src/main/java/org/datavec/api/transform/Transform.java
```diff
@@ -21,7 +21,9 @@
 import org.datavec.api.transform.ndarray.NDArrayMathFunctionTransform;
 import org.datavec.api.transform.ndarray.NDArrayScalarOpTransform;
 import org.datavec.api.transform.sequence.ReduceSequenceTransform;
+import org.datavec.api.transform.sequence.nlp.TextToIntegerSequenceTransform;
```
huitseeker: unused import ?

turambar: This is an artifact of my accidentally branching off of @AlexDBlack's branch instead of master. I suggest merging his branch before this one.

turambar:
@huitseeker Nope. Link? But several of these are NDArray specific, and the rest are so simple, I'm not sure it's worth introducing the external dependency...? @AlexDBlack requested I add these, so I'd be interested in hearing his take as well.

turambar:
Replaced by https://github.com/deeplearning4j/DataVec/pull/340.

PR #339, opened 2017-06-30
Consistent URI String formatting for FileRecordReader. Move to iterator of iterators.

knightcode:
The potentially breaking change is the addition of a requirement that 'uriStrings' always have a well formatted URIs, starting with 'file://', as it's name suggests. I updated the tests to conform to this.

The first pass through all the files in doInitialize is also wasted effort tbh and takes up memory. This moves FileRecordReader to iterate over a set of iterators. This only affects subclasses that access the internal iterator directly. All subclasses that are in any of the tests have been updated accordingly. 

## How was this patch tested?
existing maven tests

(My apologies if my first pull request here doesn't meet the standards)

### DIFF
--- a/datavec-api/src/main/java/org/datavec/api/records/reader/impl/FileRecordReader.java
+++ b/datavec-api/src/main/java/org/datavec/api/records/reader/impl/FileRecordReader.java
@@ -56,45 +56,21 @@ public class FileRecordReader extends BaseRecordReader {
 
 
     protected void doInitialize(InputSplit split) {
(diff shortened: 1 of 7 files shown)


saudet:
@crockpotveggies Does this not break any of your code?

crockpotveggies:
Much of my pre-existing work was on ImageRecordReader, not FileRecordReader. Unless we've had some changes here due to internal Transforms in datavec, I believe this should be okay. Where I'd encourage further testing is the use case where appending labels is necessary.

PR #352, opened 2017-07-13
Rewrite of SVMLight and LibSVM Readers and Writers

turambar:
Total rewrite of SVMLight and LibSVM Readers and Writers.

- Handles commented out lines
- Supports multiple labels and multilabel classification labels
- Full support for zero- and one-based indexing
- Configurable feature and label indeces in Writer
- LibSvmRecordReader is now a direct subclass of SVMLightRecordReader and has identical functionality

## How was this patch tested?

Extensive unit tests for a variety of cases and configurations:

- SVMLightRecordWriter
- LibSvmRecordWriter
- SVMLightRecordReader
- LibSvmRecordReader

and passes all previous unit tests with minor configuration changes:

- SVMRecordWriterTest
- LibSvmTest

Previous tests have been marked deprecated.

Please review


datavec-api/src/main/java/org/datavec/api/records/writer/impl/misc/SVMLightRecordWriter.java
```diff
@@ -49,50 +94,142 @@ public SVMLightRecordWriter(File path, boolean append) throws FileNotFoundExcept
...
+    }
+
+    /**
+     * Set DataVec configuration
+     *
+     * @param conf
+     */
+    @Override
+    public void setConf(Configuration conf) {
+//        super.setConf(conf); // this is not safe -- overwrites output file path!
```
agibsonccc: Maybe if we're doing things with files we could have an override? I really don't want to break the basic assumption that super can be called.

I feel like there is something better we could be doing here.

datavec-api/src/test/java/org/datavec/api/records/writer/impl/LibSvmRecordWriterTest.java
```diff
@@ -0,0 +1,326 @@
...
+
+    public static void executeTest(Configuration configWriter, Configuration configReader, File inputFile) throws Exception {
+        File tempFile = File.createTempFile("LibSvmRecordWriter", ".txt");
+        tempFile.setWritable(true);
+        tempFile.deleteOnExit();
+        if (tempFile.exists())
+            tempFile.delete();
+
+        LibSvmRecordWriter writer = new LibSvmRecordWriter(tempFile, true);
+        writer.setConf(configWriter);
```
agibsonccc: Maybe assert config values too?

datavec-api/src/test/java/org/datavec/api/records/writer/impl/LibSvmRecordWriterTest.java
```diff
@@ -0,0 +1,326 @@
...
+        File tempFile = File.createTempFile("LibSvmRecordWriter", ".txt");
+        tempFile.setWritable(true);
+        tempFile.deleteOnExit();
+        if (tempFile.exists())
+            tempFile.delete();
+
+        String lineOriginal = "13.0,14.0,15.0,4 1:1.0 2:11.0 3:12.0 4:2.0 5:3.0";
+
+        LibSvmRecordWriter writer = new LibSvmRecordWriter(tempFile, true);
+        Configuration configWriter = new Configuration();
```
agibsonccc: Maybe use try/with resources here?

datavec-api/src/test/java/org/datavec/api/records/writer/impl/LibSvmRecordWriterTest.java
```diff
@@ -0,0 +1,326 @@
...
+                new DoubleWritable(1));
+        File tempFile = File.createTempFile("LibSvmRecordWriter", ".txt");
+        tempFile.setWritable(true);
+        tempFile.deleteOnExit();
+        if (tempFile.exists())
+            tempFile.delete();
+
+        String lineOriginal = "1,3 0:1.0 1:11.0 2:12.0 3:2.0 4:3.0";
+
+        LibSvmRecordWriter writer = new LibSvmRecordWriter(tempFile, true);
```
agibsonccc: Same thing: try/with. Makes it a bit cleaner.

datavec-api/src/main/java/org/datavec/api/records/writer/impl/misc/SVMLightRecordWriter.java
AlexDBlack: Yeah, calling super.setConf(conf) should be made safe.
If it's overwriting incorrectly, then maybe set the file in the conf (if it's null) before calling super perhaps?

turambar:
@AlexDBlack @agibsonccc Seem my proposed changes in FileRecordWriter here: https://github.com/deeplearning4j/DataVec/pull/352/files#diff-7c92e92f2f4ff7f18802acfaa7183176R108

What I propose is doing checks inside of `FileRecordWriter.setConf()` to make sure we don't inadvertantly overwrite existing file path settings set by the constructor (or previous call to setConf). There are three scenarios:

- we never set file path (`writeTo` field), so we use the config settings
- config settings match existing settings, do nothing
- config settings do NOT match existing settings, throw Exception

With this, any subclass of FileRecordWriter can safely call `setConf()`.

datavec-api/src/main/java/org/datavec/api/records/writer/impl/misc/SVMLightRecordWriter.java
turambar: See recent push with my proposed fix (makes changes to superclass FileRecordWriter).

turambar:
@agibsonccc Can you confirm that I used try-with-resources in the manner you expected here: https://github.com/deeplearning4j/DataVec/pull/352/files#diff-9fd5232fb40677a882c9e1fd864d5be5R182

Assuming so, I will proceed with updating the remaining unit tests to do the same.

datavec-api/src/test/java/org/datavec/api/records/writer/impl/LibSvmRecordWriterTest.java
turambar: @agibsonccc not sure I follow -- you want me to add assertions to make sure that calling `setConf()` sets fields, etc., as expected?

datavec-api/src/main/java/org/datavec/api/records/reader/impl/misc/SVMLightRecordReader.java
```diff
@@ -1,172 +1,304 @@
...
+            } else {
+                String[] labelTokens = tokens[0].split(LABEL_DELIMITER);
+                if (numLabels < 0)
+                    numLabels = labelTokens.length;
+                if (labelTokens.length != numLabels)
+                    throw new IndexOutOfBoundsException("Found " + labelTokens.length + " labels in record, expected " + numLabels);
+                for (int i = 0; i < labelTokens.length; i++) {
+                    try { // Encode label as integer, if possible
+                        labels.add(new IntWritable(Integer.parseInt(labelTokens[i])));
+                    } catch (Exception e) {
```
saudet: Could this catch be narrowed down to NumberFormatException?

datavec-api/src/main/java/org/datavec/api/records/reader/impl/misc/SVMLightRecordReader.java
```diff
@@ -1,172 +1,304 @@
...
+
+    @Override
+    public void reset() {
+        super.reset();
+        recordLookahead = null;
+    }
+
+    @Override
+    protected void onLocationOpen(URI location) {
+        recordLookahead = null;
```
saudet: super call missing? Not that it does anything for now...

datavec-api/src/main/java/org/datavec/api/records/writer/impl/FileRecordWriter.java
```diff
@@ -105,14 +105,27 @@ public void close() {
...
+            this.append = conf.getBoolean(APPEND, true);
+            this.out = null;
+        } else {
+            String currPath = this.writeTo.getAbsolutePath();
+            String configPath = conf.get(PATH, currPath);
+            if (!configPath.equals(currPath))
+                throw new IllegalArgumentException("File path in configuration (" + configPath + ") does not match existing file path (" + currPath);
+            boolean configAppend = conf.getBoolean(APPEND, this.append);
+            if (configAppend != this.append)
+                throw new IllegalArgumentException("File append setting in configuration (" + configAppend + ") does not match existing setting (" + this.append);
```
saudet: What's the reason to not allow configuration change here?

saudet:
@turambar your try-with-resources looks good there

agibsonccc:
Could we get this in for the release?

datavec-api/src/main/java/org/datavec/api/records/writer/impl/FileRecordWriter.java
turambar: @saudet FileRecordWriter (unlike its reader counterpart) has a constructor that sets the file output path and creates the output stream. Previous behavior was that `setConf()` overwrote this setting without any kind of warning or error. Probably a more comprehensive fix would either remove that constructor or these specific configuration options, but either of those decisions would likely cause a domino effect impacting other parts of DataVec. Given that we're already planning an overhaul of DataVec anyway, I'd like to defer that kind of decision. For now, the short term fix is to do some checks inside of `setConf()` to verify that a non-default output path isn't being obliterated by `setConf()`.

turambar:
@agibsonccc @AlexDBlack @saudet OK I made all of the requested changes:

- use try-with-resources inside unit tests
- resolve the conflict between the FileRecordWriter constructor and `setConf()`
- change Exception to NumberFormatException
- added call to super inside `OnLocationOpen()`

All relevant unit tests, including legacy ones, pass. I also merged with master in a second branch, rebuilt DataVec, and ran all the DataVec tests. Everything appears in order. This should be fine to include in the upcoming release (if time allows), but I'll leave that up to you guys. Let me know what you think!

PR #356, opened 2017-07-15
Inferred schema, CSV parser, and more transforms.

crockpotveggies:

To make CSV parsing more user-friendly, I've added some functionality that allows "magic" inference of schema from a sample CSV file. To assist transforms of fields marked as strings, I've added `convertToXnumberX` functions to TransformProcess. Also, CSV parsing was not working well for me, so I've added OpenCSV to the CSVRecordReader and made it serializable.

## How was this patch tested?

Ran CSV tests as well as tried this in my own project on Spark local.

### DIFF
--- a/datavec-api/pom.xml
+++ b/datavec-api/pom.xml
@@ -128,6 +128,13 @@
             <version>${nd4j.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <!-- csv parser, same dep used by spark -->
+        <dependency>
+            <groupId>net.sf.opencsv</groupId>
+            <artifactId>opencsv</artifactId>
+            <version>2.3</version>
+        </dependency>
     </dependencies>
 
     <profiles>
--- a/datavec-api/src/main/java/org/datavec/api/conf/Configuration.java
+++ b/datavec-api/src/main/java/org/datavec/api/conf/Configuration.java
@@ -389,6 +389,36 @@ public class Configuration implements Iterable<Map.Entry<String, String>>, Writa
     }
 
     /**
+     * Get the char value of the <code>name</code> property, <code>null</code> if
+     * no such property exists.
+     *
+     * Values are processed for <a href="#VariableExpansion">variable expansion</a>
+     * before being returned.
+     *
+     * @param name the property name.
+     * @return the value of the <code>name</code> property,
+     *         or null if no such property exists.
+     */
+    public char getChar(String name) {
+        return getProps().getProperty(name).charAt(0);
+    }
+
+    /**
+     * Get the char value of the <code>name</code> property, <code>null</code> if
+     * no such property exists.
+     *
+     * Values are processed for <a href="#VariableExpansion">variable expansion</a>
+     * before being returned.
+     *
+     * @param name the property name.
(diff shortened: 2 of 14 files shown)


datavec-api/pom.xml
```diff
@@ -128,6 +128,13 @@
             <version>${nd4j.version}</version>
             <scope>test</scope>
         </dependency>
+
+        <!-- csv parser, same dep used by spark -->
```
huitseeker: Well, yes and no (as we've discussed, they really "use" univocity).

You may want to indicate what is bringing in the dependency, esp since opencsv is slower:
https://github.com/uniVocity/csv-parsers-comparison

But I agree this improves on the codebase.

datavec-api/src/main/java/org/datavec/api/conf/Configuration.java
```diff
@@ -389,6 +389,36 @@ public String getRaw(String name) {
...
+     * no such property exists.
+     *
+     * Values are processed for <a href="#VariableExpansion">variable expansion</a>
+     * before being returned.
+     *
+     * @param name the property name.
+     * @return the value of the <code>name</code> property,
+     *         or null if no such property exists.
+     */
+    public char getChar(String name) {
```
huitseeker: You may want to rename this `getCharFromProperty`, the natural expectation for a `getChar` function being some `char getChar(s: String)`.

datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVRecordReader.java
```diff
@@ -40,19 +40,16 @@
...
 public class CSVRecordReader extends LineRecordReader {
-    /** A regex delimiter that can parse quotes (string literals) that may have commas in them: http://stackoverflow.com/a/1757107/523744
-     * Note: This adds considerable overhead compared to the default "," delimiter, and should only be used when necessary.
-     * */
-    public final static String QUOTE_HANDLING_DELIMITER = ",(?=([^\"]*\"[^\"]*\")*[^\"]*$)";
     private boolean skippedLines = false;
     protected int skipNumLines = 0;
-    protected String delimiter = DEFAULT_DELIMITER;
-    protected String quote = null;
-    public final static String DEFAULT_DELIMITER = ",";
```
huitseeker: Is there any chance any upstream code would depend on this modified public member? in dl4j-examples perhaps ?

datavec-api/src/main/java/org/datavec/api/transform/transform/doubletransform/ConvertToDouble.java
```diff
@@ -0,0 +1,57 @@
...
+import org.datavec.api.writable.IntWritable;
+import org.datavec.api.writable.Writable;
+
+/**
+ * Convert any value to an Integer.
+ *
+ * @author Justin Long (crockpotveggies)
+ */
+@NoArgsConstructor
+public class ConvertToDouble extends BaseDoubleTransform {
```
huitseeker: Oooh, nice, could we get floats too ?

PR #363, opened 2017-07-25
Re-add missing CSVRecordReader constructor (with string delim and quote).

turambar:
# What changes were proposed in this pull request?

It's all in the title. @crockpotveggies broke it.

## How was this patch tested?

No tests -- but helps maintain backwards compatibility.

### DIFF
--- a/datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVRecordReader.java
+++ b/datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVRecordReader.java
@@ -103,7 +103,10 @@ public class CSVRecordReader extends LineRecordReader {
      * @param skipNumLines the number of lines to skip
      * @param delimiter the delimiter
      * @param quote the quote to strip
+     * @deprecated This constructor is deprecated; use {@link #CSVRecordReader(int, char)} or
+     * {@link #CSVRecordReader(int, char, char)}
      */
+    @Deprecated
     public CSVRecordReader(int skipNumLines, String delimiter, String quote) {
         this(skipNumLines, stringDelimToChar(delimiter), stringDelimToChar(quote));
     }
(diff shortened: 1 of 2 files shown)


turambar:
@crockpotveggies

datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVRecordReader.java
```diff
@@ -98,6 +98,16 @@ public CSVRecordReader(int skipNumLines, char delimiter, char quote) {
         this.csvParser = new SerializableCSVParser(delimiter, quote);
     }
 
+    /**
+     * Skip lines, use delimiter, and strip quotes
+     * @param skipNumLines the number of lines to skip
+     * @param delimiter the delimiter
+     * @param quote the quote to strip
+     */
+    public CSVRecordReader(int skipNumLines, String delimiter, String quote) {
```
huitseeker: This ignores `skipNumLines` and probably should not.

turambar: Fixed.

crockpotveggies:
This has already been addressed by https://github.com/deeplearning4j/DataVec/pull/360/files I don't understand what's different...?

turambar:
Adding @AlexDBlack so he can sort out the overlap with the other PR.

crockpotveggies:
Alex's PR has already been merged. Are you on an out of date branch? Or are you adding an entirely new constructor?

AlexDBlack:
LGTM, except that I'd probably prefer an ```@deprecated``` on the added constructor... given that only characters are really supported now (not strings), the char delim should be the main constructor IMO. FYI my earlier PR was a different constructor signature (only delim, this one has both delim and quote)

Issue #362, opened 2017-07-25
Add a JDBC record reader


dalsh:
Proposal for a new record reader to iterate on a jdbc ResultSet and transform rows into record values.

A few remarks and questions with no special order :
- Commons db offers some abstractions but the standard jdbc api may offer more control (in particular concerning paging). JOOQ has interesting features too, like dialect agnostic request generation DSL
- Minimum required parameters should be a datasource and a query
- Any opinion on connection persistence and (potential) pooling ?
- Should this be a separate maven module ?
- How to map relational column types to instances of Writables ? I've seen some related stuff in the interface layer between local and tablesaw's fork in dataframe ([SqlResultSetReader](https://github.com/deeplearning4j/DataVec/blob/master/datavec-dataframe/src/main/java/org/datavec/dataframe/io/jdbc/SqlResultSetReader.java) in this same module seems to be doing some mapping as well)

agibsonccc:
We use commonsdb..jooq is a bit heavy for this but I wouldn't mind using it if done right.

DBUtils/commons db makes more sense

No opinions on pooling.

Yes it should be a separate module.

Writables are generally relative to type. I would look at what the csv record reader does and immitate that.

dalsh:
Thanks for your answers. Concerning the Writables, I meant that CSVRecordReader is always reading Texts, while the JDBCRecordReader may have to read other types of Writable and some mapping should occur.

Can you assign this to me ? Started working on it here https://github.com/dalsh/DataVec/tree/features/jdbc-record-reader-362

agibsonccc:
yes you would need different Writable types: https://github.com/deeplearning4j/DataVec/blob/master/datavec-api/src/test/java/org/datavec/api/records/reader/impl/transform/TransformProcessRecordReaderTests.java#L29

dalsh:
I opened PR #364 so you can start early reviewing.

I have 2 questions :
- what should metadata objects store in order to re-fetch data ? One idea could be to ask the user to provide a primary key column name and a table name so we can store the related key value and table name in the metadata, but that only covers simple cases (if you iterate over a join or select in select request this won't work)

- what's an inputsplit exactly and how should the concept be applied to JDBC record reading ?

agibsonccc:
input splits for data base could be queries or even batched results

for refetching data should be able to reuse the same query I'd imagine?

dalsh:
[...]

PR #364, opened 2017-07-27
Add JDBC record reader (#362)

dalsh:

This adds the JDBC record reader mentioned in #362 . Work is still in progress but I'm putting it here if you want to start reviewing it anyway.

## How was this patch tested?

This should be fully covered in JDBCRecordReaderTest

### DIFF
--- a/datavec-jdbc/src/main/java/org/datavec/api/records/reader/impl/jdbc/JDBCRecordReader.java
+++ b/datavec-jdbc/src/main/java/org/datavec/api/records/reader/impl/jdbc/JDBCRecordReader.java
@@ -1,0 +1,92 @@
+package org.datavec.api.records.reader.impl.jdbc;
+
+import org.datavec.api.conf.Configuration;
+import org.datavec.api.records.Record;
+import org.datavec.api.records.metadata.RecordMetaData;
+import org.datavec.api.records.reader.BaseRecordReader;
+import org.datavec.api.split.InputSplit;
+import org.datavec.api.writable.Writable;
+
+import javax.sql.DataSource;
+import java.io.DataInputStream;
+import java.io.IOException;
+import java.net.URI;
+import java.sql.ResultSet;
+import java.util.List;
+
+public class JDBCRecordReader extends BaseRecordReader {
+
+    private DataSource dataSource;
+    private ResultSet resultSet;
+    private String query;
+
+    public JDBCRecordReader(DataSource dataSource, String query) {
+        this.dataSource = dataSource;
+        this.query = query;
+    }
+
+    @Override
+    public void initialize(InputSplit split) throws IOException, InterruptedException {
+
+    }
+
+    @Override
+    public void initialize(Configuration conf, InputSplit split) throws IOException, InterruptedException {
+
+    }
+
+    @Override
+    public List<Writable> next() {
+        return null;
+    }
+
+    @Override
+    public boolean hasNext() {
+        return false;
+    }
+
+    @Override
+    public List<String> getLabels() {
+        return null;
+    }
+
+    @Override
+    public void reset() {
+
+    }
+
+    @Override
+    public List<Writable> record(URI uri, DataInputStream dataInputStream) throws IOException {
+        return null;
+    }
+
+    @Override
+    public Record nextRecord() {
+        return null;
+    }
+
+    @Override
+    public Record loadFromMetaData(RecordMetaData recordMetaData) throws IOException {
+        return null;
+    }
+
+    @Override
+    public List<Record> loadFromMetaData(List<RecordMetaData> recordMetaDatas) throws IOException {
+        return null;
+    }
+
+    @Override
+    public void close() throws IOException {
+
+    }
+
+    @Override
+    public void setConf(Configuration conf) {
+
+    }
+
+    @Override
+    public Configuration getConf() {
+        return null;
+    }
+}
--- a/datavec-jdbc/src/main/java/org/datavec/api/records/reader/impl/jdbc/JDBCRecordReader.java
+++ b/datavec-jdbc/src/main/java/org/datavec/api/records/reader/impl/jdbc/JDBCRecordReader.java
@@ -1,5 +1,6 @@
 package org.datavec.api.records.reader.impl.jdbc;
 
+import com.zaxxer.hikari.util.DriverDataSource;
 import java.io.DataInputStream;
 import java.io.IOException;
 import java.net.URI;
@@ -10,6 +11,7 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Properties;
 import javax.sql.DataSource;
 import lombok.Setter;
 import org.apache.commons.dbutils.DbUtils;
@@ -29,8 +31,8 @@ import org.datavec.api.writable.Writable;
  */
 public class JDBCRecordReader extends BaseRecordReader {
 
-    private final DataSource dataSource;
     private final String query;
+    private DataSource dataSource;
     private Connection conn;
     private Statement statement;
     private ResettableResultSetIterator iter;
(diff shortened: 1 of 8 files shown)


datavec-jdbc/src/main/java/org/datavec/api/records/reader/impl/jdbc/JDBCRecordReader.java
```diff
@@ -0,0 +1,145 @@
...
+            this.meta = rs.getMetaData();
+            this.iter = new ResettableResultSetIterator(rs);
+        } catch (SQLException e) {
+            closeJdbc();
+            throw new RuntimeException("Could not connect to the database", e);
+        }
+    }
+
+    @Override
+    public void initialize(Configuration conf, InputSplit split) throws IOException, InterruptedException {
```
agibsonccc: Shouldn't we allow data source configuration from the strings/conf here.

datavec-jdbc/src/main/java/org/datavec/api/records/reader/impl/jdbc/JDBCRecordReader.java
```diff
@@ -0,0 +1,145 @@
...
+                try {
+                    Object columnValue = next[i];
+                    if (trimStrings && columnValue instanceof String) {
+                        columnValue = ((String) columnValue).trim();
+                    }
+                    // Note, getColumnType first argument is column number starting from 1
+                    Writable writable = JdbcWritableConverter.convert(columnValue, meta.getColumnType(i + 1));
+                    ret.add(writable);
+                } catch (SQLException e) {
+                    closeJdbc();
```
agibsonccc: try/with resource clauses here instead?

datavec-jdbc/src/main/java/org/datavec/api/records/reader/impl/jdbc/JDBCRecordReader.java
```diff
@@ -0,0 +1,145 @@
...
+        return null;
+    }
+
+    @Override
+    public void reset() {
+        iter.reset();
+    }
+
+    @Override
+    public List<Writable> record(URI uri, DataInputStream dataInputStream) throws IOException {
```
agibsonccc: throw an UnsupportedOperationExecption instead of returning null

datavec-jdbc/src/main/java/org/datavec/api/records/reader/impl/jdbc/JDBCRecordReader.java
```diff
@@ -0,0 +1,145 @@
...
+        return null;
+    }
+
+    @Override
+    public Record nextRecord() {
```
agibsonccc: Record shouldn't return null here. Please add the proper return and assoiated test there.

datavec-jdbc/src/main/java/org/datavec/api/records/reader/impl/jdbc/JDBCRecordReader.java
```diff
@@ -0,0 +1,145 @@
...
+        return null;
+    }
+
+    @Override
+    public Record loadFromMetaData(RecordMetaData recordMetaData) throws IOException {
```
agibsonccc: RecordMetaData and jdbc schemas should map. I feel like there should be a conversion here.

datavec-jdbc/src/main/java/org/datavec/api/records/reader/impl/jdbc/JDBCRecordReader.java
```diff
@@ -0,0 +1,145 @@
...
+        return null;
+    }
+
+    @Override
+    public List<Record> loadFromMetaData(List<RecordMetaData> recordMetaDatas) throws IOException {
+        return null;
```
agibsonccc: Same.

datavec-jdbc/src/main/java/org/datavec/api/records/reader/impl/jdbc/JDBCRecordReader.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    private void closeJdbc() {
+        DbUtils.closeQuietly(statement);
+        DbUtils.closeQuietly(conn);
+    }
+
+    @Override
+    public void setConf(Configuration conf) {
+
```
agibsonccc: setConf/ and conf() should return something.

datavec-jdbc/src/main/java/org/datavec/api/records/reader/impl/jdbc/JDBCRecordReader.java
dalsh: To deal with that I've added HikariCP as dependency for their DriverDataSource, is that ok ?

datavec-jdbc/src/main/java/org/datavec/api/records/reader/impl/jdbc/JDBCRecordReader.java
dalsh: Wouldn't that close at the end of the block in any case ? I think it may be better to leave the connection open unless something goes wrong, and let the user close it when he's finished reading .. ?

dalsh:
I pushed some progress, I'm going to look into InputSplits now

datavec-jdbc/src/main/java/org/datavec/api/util/jdbc/JdbcWritableConverter.java
```diff
@@ -0,0 +1,57 @@
...
+            case Types.REAL:
+                return new FloatWritable((float) columnValue);
+
+            case Types.DOUBLE:
+                return new DoubleWritable((double) columnValue);
+
+            case Types.INTEGER:
+            case Types.SMALLINT:
+            case Types.TINYINT:
+            case Types.BIGINT:
```
AlexDBlack: BIGINT is equivalent to long I believe? Probably want to return a LongWritable.

datavec-jdbc/src/main/java/org/datavec/api/util/jdbc/JdbcWritableConverter.java
```diff
@@ -0,0 +1,57 @@
...
+            case Types.NCHAR:
+            case Types.NVARCHAR:
+            case Types.VARCHAR:
+                return new Text((String) columnValue);
+
+            case Types.DECIMAL:
+            case Types.FLOAT:
+            case Types.NUMERIC:
+            case Types.REAL:
+                return new FloatWritable((float) columnValue);
```
AlexDBlack: Are we likely to run into overflow issues here? I think float/real should be ok, but decimal/numerical could in principle overflow float min/max?

dalsh: I had a similar thought recently .. Wouldn't we need something like a BigDecimalWritable for this case ?

AlexDBlack: DoubleWritable would be a step up from FloatWritable, and should be fine for 99.9% of users. I'm hesitant to add another writable just for that possibility, at this point.

dalsh:
I've done the requested changes, added documentation, some fixes after self review and rebased on master. This is starting to look good to me too so I removed the WIP tag.

datavec-jdbc/src/test/java/org/datavec/api/records/reader/impl/TestDb.java
```diff
@@ -0,0 +1,206 @@
+package org.datavec.api.records.reader.impl;
```
huitseeker: This is copy-paste from Tablesaw. Is tablesaw's test-jar usable in this context ?
https://github.com/lwhite1/tablesaw/blob/master/src/test/java/com/github/lwhite1/tablesaw/util/TestDb.java

dalsh: Ah yes, this slipped my mind. I used this for convenience but I intended to replace it with more sophisticated data tables. I'll work on that and also create some tests on the data mapping which I intended to do too.

dalsh: I added some unit tests which revealed some flaws in the data mapping. It's now fixed

agibsonccc:
Merging due to lack of activity. Thanks for the contribution!

PR #369, opened 2017-08-04
Version check support

AlexDBlack:
https://github.com/deeplearning4j/nd4j/pull/2016


huitseeker:
Oh, maybe I know why this is the case: when you have a pom-only project, such as datavec-xxx-parent, there is no jar to add to despite a git.properties still being created with this method. If that's what you're witnessing, it shouldn't be a problem ?

AlexDBlack:
@huitseeker They are being generated correctly - but they don't end up in the JAR files after an install etc. It's probably just maven config somewhere - I'll look into it.

PR #372, opened 2017-08-05
Introduce SKIP_BUILD mode to perform-release.sh script

saudet:

Allow updating the source code for a release before making the final build on the "latest_release" tag.

## How was this patch tested?

Build succeeds with source code modified by the script, with no reference to SNAPSHOT artifacts.

### DIFF
--- a/perform-release.sh
+++ b/perform-release.sh
@@ -9,9 +9,7 @@ fi
 RELEASE_VERSION=$1
 SNAPSHOT_VERSION=$2
 STAGING_REPOSITORY=$3
-if [[ -z ${SKIP_BUILD:-} ]]; then
-    SKIP_BUILD=0
-fi
+SKIP_BUILD=${SKIP_BUILD:-0}
 
 echo "Releasing version $RELEASE_VERSION ($SNAPSHOT_VERSION) to repository $STAGING_REPOSITORY"
 echo "========================================================================================"
--- a/perform-release.sh
+++ b/perform-release.sh
@@ -9,6 +9,9 @@ fi
 RELEASE_VERSION=$1
 SNAPSHOT_VERSION=$2
 STAGING_REPOSITORY=$3
+if [[ -z ${SKIP_BUILD:-} ]]; then
+    SKIP_BUILD=0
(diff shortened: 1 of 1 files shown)


perform-release.sh
```diff
@@ -25,17 +28,20 @@ for f in $(find . -name 'pom.xml' -not -path '*target*'); do
 done
 mvn versions:set -DallowSnapshots=true -DgenerateBackupPoms=false -DnewVersion=$RELEASE_VERSION
 
-mvn clean deploy -Dgpg.executable=gpg2 -DperformRelease -Psonatype-oss-release -DskipTests -DstagingRepositoryId=$STAGING_REPOSITORY
-source change-spark-versions.sh 2
-mvn clean deploy -Dgpg.executable=gpg2 -DperformRelease -Psonatype-oss-release -DskipTests -DstagingRepositoryId=$STAGING_REPOSITORY
-source change-scala-versions.sh 2.10
-source change-spark-versions.sh 1
-mvn clean deploy -Dgpg.executable=gpg2 -DperformRelease -Psonatype-oss-release -DskipTests -DstagingRepositoryId=$STAGING_REPOSITORY
+if [[ "${SKIP_BUILD}" == "0" ]]; then
```
huitseeker: This is OK, but it assumes the current state of the repo is Scala 2.11 / Spark 1, which may or may not be true (see https://github.com/deeplearning4j/DataVec/commit/485fd7c3725c4172d3ed4a736e08d34df538fe4b ).

This should really call buildmultiplescalaversions.sh. The script will happily use all these arguments, and then we only have one place in which we have to maintain the versions-switching.

perform-release.sh
```diff
@@ -9,6 +9,9 @@ fi
 RELEASE_VERSION=$1
 SNAPSHOT_VERSION=$2
 STAGING_REPOSITORY=$3
+if [[ -z ${SKIP_BUILD:-} ]]; then
```
huitseeker: Nitpick : `SKIP_BUILD=${SKIP_BUILD:-0}`

PR #373, opened 2017-08-06
Minor fixes for SVMLight and LibSvm record readers and writers

turambar:

(1) added support for records with no labels, either fully unlabeled data or multilabel with no positive labels
(2) for features and sparse multilabels, now pre-allocate List<Writables> with Collections.numCopies() and call set(index, value) to update features, rather than using ad hoc while loops.

Added new unit tests or updated existing unit tests to cover various "no label" scenarios. All unit tests (including legacy tests) pass.

## How was this patch tested?

Please review


turambar:
TL;DR previously, support for records with no labels (either fully unlabeled data or multilabel records with no positive labels) was incomplete. Now we cover most common scenarios, and we're now consistent with scikit-learn's functionality and behavior.

PR #374, opened 2017-08-07
Fix buildmultiplescalaversions.sh for the Mac

huitseeker:
No readlink -f available there without coreutils


huitseeker:
Approval derived from https://github.com/deeplearning4j/nd4j/pull/2024

PR #380, opened 2017-08-10
Align dataframe filter naming

benmccann:

We have two methods `isIn` and `IsInSet` that are exactly the same, but have different names. This gives them the same name. This also aligns the naming with naming that is used elsewhere.

This also renames `StringInSet` to `StringIsIn` to be analagous to `IntIsIn`.

Both of these changes align this repo with the upstream.

## How was this patch tested?

Ran `DataVec/datavec-dataframe $ mvn test`


benmccann:
Good call. Done

benmccann:
Closing in favor of unfork

PR #381, opened 2017-08-10
Remove shade plugin from dataframe

benmccann:
Using the shade plugin can cause problems when loading from Maven/Gradle/SBT/etc.  https://github.com/jtablesaw/tablesaw/issues/108


benmccann:
I'm going to close this PR in favor of removing datavec-dataframe and replacing with the upstream tablesaw. I'll leave my other PR open as it will help with that

PR #382, opened 2017-08-10
Make graphical tests return in CI environments

huitseeker:
See also http://www.oracle.com/technetwork/articles/javase/headless-136834.html


agibsonccc:
@huitseeker we might want to bake this internally somehow if we can..does this depend on a particular jdk?

huitseeker:
@agibsonccc No, all Java projects I know use `-Djava.awt.headless=true` in JVM options for running tests in CI, and all projects I know test for this option in the way I use in this PR.

PR #398, opened 2017-08-18
Add variable sliding window CSV record reader

crockpotveggies:

Adds support for sliding windows across CSV files. Windows are variable in length, according to index being read. User can also set "stride" of window.

## How was this patch tested?

See added tests. Currently passing.

### DIFF
--- a/datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVVariableSlidingWindowRecordReader.java
+++ b/datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVVariableSlidingWindowRecordReader.java
@@ -35,6 +35,9 @@ import java.util.*;
 /**
  * A sliding window of variable size across an entire CSV.
  *
+ * In practice the sliding window size starts at 1, then linearly increase to maxLinesPer sequence, then
+ * linearly decrease back to 1.
+ *
  * @author Justin Long (crockpotveggies)
  */
 public class CSVVariableSlidingWindowRecordReader extends CSVRecordReader implements SequenceRecordReader {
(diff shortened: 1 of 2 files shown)


datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVVariableSlidingWindowRecordReader.java
```diff
@@ -0,0 +1,235 @@
...
+import org.nd4j.linalg.primitives.Triple;
+
+import java.io.DataInputStream;
+import java.io.IOException;
+import java.net.URI;
+import java.util.*;
+
+/**
+ * A sliding window of variable size across an entire CSV.
+ *
```
AlexDBlack: Can we get a note on how the windowing works for the edges (start + end)?
From what I understand, it'll start with 1, then linearly increase to maxLinesPer sequence, then linearly decrease back to 1.

datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVVariableSlidingWindowRecordReader.java
```diff
@@ -0,0 +1,235 @@
...
+        this(maxLinesPerSequence, 0, stride, String.valueOf(CSVRecordReader.DEFAULT_DELIMITER));
+    }
+
+    /**
+     *
+     * @param maxLinesPerSequence    Number of lines in each sequences
+     * @param skipNumLines         Number of lines to skip at the start of the file (only skipped once, not per sequence)
+     * @param delimiter            Delimiter between entries in the same line, for example ","
+     */
+    public CSVVariableSlidingWindowRecordReader(int maxLinesPerSequence, int skipNumLines, int stride, String delimiter) {
```
AlexDBlack: Need Stride arg javadoc. Also probably want check on stride - make sure it's > 0

datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVVariableSlidingWindowRecordReader.java
```diff
@@ -0,0 +1,235 @@
...
+
+            if (queue.size() > maxLinesPerSequence || exhausted)
+                queue.poll();
+        }
+
+        for(List<Writable> line : queue) {
+            sequence.add(line);
+        }
+
+        Collections.reverse(sequence);
```
AlexDBlack: Why reverse the sequence? This doesn't seem correct to me...
(And if it is: I'd use queue.addFirst, then just do sequence = new ArrayList(queue)), insthead of the for loop + reverse)

datavec-api/src/test/java/org/datavec/api/records/reader/impl/CSVVariableSlidingWindowRecordReaderTest.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Tests for variable sliding window csv reader.
+ *
+ * @author Justin Long (crockpotveggies)
+ */
+public class CSVVariableSlidingWindowRecordReaderTest {
```
AlexDBlack: Need better tests: notably,
(a) assert total number of sequences returned
(b) Test the metadata functionality

crockpotveggies: Fixed, and added test for strides.

datavec-api/src/main/java/org/datavec/api/records/reader/impl/csv/CSVVariableSlidingWindowRecordReader.java
crockpotveggies: Fixed.

Issue #400, opened 2017-08-21
Failing test for CSVSparkTransformTest.testSingleBatchSequence


huitseeker:
#### Issue Description

```
Running org.datavec.spark.transform.CSVSparkTransformTest                      
Base 64ed array [[1.00,  2.00],                                                
 [1.00,  2.00],  
 [1.00,  2.00]]                                                                
SequenceBatchCSVRecord(records=[[BatchCSVRecord(records=[SingleCSVRecord(values=[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]), SingleCSVRecord(values=[0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]), SingleCSVRecord(values=[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]), SingleCSVRecord(values=[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0])])], [BatchCSVRecord(records=[SingleCSVRecord(values=[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0
, 0, 0, 0, 0, 0, 0, 0]), SingleCSVRecord(values=[0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]), SingleCSVRecord(values=[0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 
[...]

PR #402, opened 2017-08-22
tests pass now

agibsonccc:
Fixes: https://github.com/deeplearning4j/DataVec/issues/400

## How was this patch tested?

Please review


AlexDBlack:
Whether the test is passing or not - this doesn't look right to me.

Let's assume 'strings' is a list of sequences. Then strings.size() == number of sequences.

Sequences in datavec are seq=List<List<Writable>>, with seq.size() == sequence length, and seq.get(i).size() == strings.get(j).get(i).size() == values per time step.

Time series arrays in DL4J are [minibatch, size, tsLength] which gives (strings.size(), strings.get(0).get(0).size(), strings.get(0).size()), assuming same time series length for all sequences.

So again, the original (pre-change) version looks correct to me. The only other thing I can think of is that the input data (strings value) is stored columnar not row-based?

agibsonccc:
Yeah ok..so maybe the test was wrong. There was a reason I didnt merge it
yet..

PR #408, opened 2017-08-28
Adding TextToTermIndexSequenceTransform

turambar:

Adding a TextToTermIndexSequenceTransform (similar to TextToCharacterIndexTransform but with terms).

## How was this patch tested?

Added a unit test to TestTransforms.

### DIFF
--- a/datavec-api/src/main/java/org/datavec/api/transform/transform/nlp/TextToTermIndexSequenceTransform.java
+++ b/datavec-api/src/main/java/org/datavec/api/transform/transform/nlp/TextToTermIndexSequenceTransform.java
@@ -38,9 +38,10 @@ import java.util.*;
  * @author Dave Kale
  */
 @Data
+//@EqualsAndHashCode()
 @EqualsAndHashCode(callSuper = true, exclude = {"writableMap"})
 @JsonIgnoreProperties({"writableMap"})
-@JsonInclude(JsonInclude.Include.ALWAYS)
+@JsonInclude(JsonInclude.Include.NON_NULL)
 public class TextToTermIndexSequenceTransform extends BaseSequenceExpansionTransform {
 
     private Map<String,Integer> wordIndexMap;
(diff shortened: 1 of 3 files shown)


datavec-api/src/main/java/org/datavec/api/transform/transform/nlp/TextToTermIndexSequenceTransform.java
```diff
@@ -0,0 +1,119 @@
...
+/**
+ *
+ * Convert each text value in a sequence to a longer sequence of integer indices.
+ * For example, "zero one two" would be converted to [0, 1, 2]. Values in other
+ * columns will be duplicated.
+ *
+ * @author Dave Kale
+ */
+@Data
+@EqualsAndHashCode(callSuper = true, exclude = {""})
```
AlexDBlack: probably exclude={"writableMap"}

datavec-api/src/main/java/org/datavec/api/transform/transform/nlp/TextToTermIndexSequenceTransform.java
```diff
@@ -0,0 +1,119 @@
...
+@JsonIgnoreProperties({"wordIndexMap", "writableMap"})
```
AlexDBlack: Remove "wordIndexMap" otherwise your transform can't be (de)serialized from/to JSON (your vocab won't be present).

turambar:
@AlexDBlack addressed feedback on JSON annotations, and added a unit test to test serialization. However, the unit test fails because it looks like the call to equals is not properly ignoring the `writableMap` field as specified in the EqualsAndHashCode annotation. Any thoughts would be much appreciated!

turambar:
@AlexDBlack: I fixed the annotations on `BaseSequenceTransform` and the new `TextToTermIndexSequenceTransform` -- unit test now passes.

PR #409, opened 2017-08-28
Add ConditionalReplaceValueTransformWithDefault

huitseeker:
A compact version of partitioning the replacement values for a column by a condition,
offshoot of work on the intrusion demo.

### DIFF
--- a/datavec-api/src/main/java/org/datavec/api/transform/TransformProcess.java
+++ b/datavec-api/src/main/java/org/datavec/api/transform/TransformProcess.java
@@ -45,6 +45,7 @@ import org.datavec.api.transform.transform.categorical.StringToCategoricalTransf
 import org.datavec.api.transform.transform.column.*;
 import org.datavec.api.transform.transform.condition.ConditionalCopyValueTransform;
 import org.datavec.api.transform.transform.condition.ConditionalReplaceValueTransform;
+import org.datavec.api.transform.transform.condition.ConditionalReplaceValueTransformWithDefault;
 import org.datavec.api.transform.transform.doubletransform.*;
 import org.datavec.api.transform.transform.integer.ConvertToInteger;
 import org.datavec.api.transform.transform.integer.IntegerColumnsMathOpTransform;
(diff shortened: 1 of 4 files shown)


huitseeker:
@AlexDBlack Indeed. I hope my latest fixes all that.

To be clear, this operation is strictly better than two successive conditional replaces, not just form performance, but
- in the case where we wish to have a non-empty intersection between the image of one side of the partition and the domain of the other (e.g. if x  <= 0 replace by 1 else replace by 0)
- and it makes the case where the condition is a pain to negate, such as regexes, a bit easier (than the use of Boolean condition operators).

datavec-api/src/main/java/org/datavec/api/transform/TransformProcess.java
```diff
@@ -1327,6 +1328,19 @@ public Builder conditionalReplaceValueTransform(String column, Writable newValue
...
         /**
+         * Replace the values in a specified column with a specified new value, if some condition holds.
+         * If the condition does not hold, the original values are not modified.
+         *
+         * @param column    Column to operate on
+         * @param newVal  Value to use as replacement, if condition is satisfied
+         * @param defaultVal  Value to use as replacement, if condition is satisfied
+         * @param condition Condition that must be satisfied for replacement
+         */
+        public Builder conditionalReplaceValueTransformWithDefault(String column, Writable newVal, Writable defaultVal, Condition condition) {
```
AlexDBlack: Should align with the ```yesVal``` and ```noVal``` semantics of the constructor.

datavec-api/src/main/java/org/datavec/api/transform/TransformProcess.java
```diff
@@ -1327,6 +1328,19 @@ public Builder conditionalReplaceValueTransform(String column, Writable newValue
         }
 
         /**
+         * Replace the values in a specified column with a specified new value, if some condition holds.
+         * If the condition does not hold, the original values are not modified.
```
AlexDBlack: Still incorrect, does not match class javadoc.

datavec-api/src/main/java/org/datavec/api/transform/TransformProcess.java
```diff
@@ -1327,6 +1328,19 @@ public Builder conditionalReplaceValueTransform(String column, Writable newValue
...
+         *
+         * @param column    Column to operate on
+         * @param newVal  Value to use as replacement, if condition is satisfied
+         * @param defaultVal  Value to use as replacement, if condition is satisfied
```
AlexDBlack: is *not* satisfied

huitseeker:
Ouch, apologies: I messed up, I should have paid more attention to this.

PR #410, opened 2017-08-29
Equality semantics for Writable types that are a Number

### LINKED ISSUE #265: Review Writable equality
```assertEquals(new DoubleWritable(3), new IntWritable(3));```
That fails :/

We should ensure that where appropriate, the writables are actually equal.
That's easy for integers etc, but we might need an epsilon or something for double/float equality.

### PR DESCRIPTION
**Pull request by huitseeker**
- ensures `hashCode` and `equals` reflects the semantics of the value for `(Byte|Int|Long|Double|Float)Writable`
- adds a facade to `fuzzyEquals` (Guava) for those types

Fixes #265

### DIFF
--- a/datavec-api/src/main/java/org/datavec/api/writable/ByteWritable.java
+++ b/datavec-api/src/main/java/org/datavec/api/writable/ByteWritable.java
@@ -17,6 +17,7 @@
 package org.datavec.api.writable;
 
 
+import com.google.common.math.DoubleMath;
 import org.datavec.api.io.WritableComparable;
 import org.datavec.api.io.WritableComparator;
 import org.nd4j.shade.jackson.annotation.JsonProperty;
@@ -57,6 +58,22 @@ public class ByteWritable implements WritableComparable {
 
     public void write(DataOutput out) throws IOException {
         out.writeByte(value);
+    }
+
+    public boolean fuzzyEquals(Writable o, double tolerance) {
+        double other;
+        if (o instanceof IntWritable){
+            other = ((IntWritable) o).toDouble();
+        } else if (o instanceof  LongWritable) {
+            other = ((LongWritable) o).toDouble();
+        } else if (o instanceof ByteWritable) {
+            other = ((ByteWritable) o).toDouble();
+        } else if (o instanceof  DoubleWritable) {
+            other = ((DoubleWritable) o).toDouble();
+        } else if (o instanceof  FloatWritable) {
+            other = ((FloatWritable) o).toDouble();
+        } else { return false; }
+        return DoubleMath.fuzzyEquals(this.value, other, tolerance);
     }
 
     /** Returns true iff <code>o</code> is a ByteWritable with the same value. */
--- a/datavec-api/src/main/java/org/datavec/api/writable/ByteWritable.java
+++ b/datavec-api/src/main/java/org/datavec/api/writable/ByteWritable.java
@@ -88,7 +88,7 @@ public class ByteWritable implements WritableComparable {
     }
 
     public int hashCode() {
-        return (int) value;
+        return (int)value;
     }
 
     /** Compares two ByteWritables. */
--- a/datavec-api/src/main/java/org/datavec/api/writable/ByteWritable.java
+++ b/datavec-api/src/main/java/org/datavec/api/writable/ByteWritable.java
@@ -61,11 +61,13 @@ public class ByteWritable implements WritableComparable {
 
     /** Returns true iff <code>o</code> is a ByteWritable with the same value. */
     public boolean equals(Object o) {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: datavec-api/src/main/java/org/datavec/api/writable/IntWritable.java [hunk 1]
```diff
@@ -59,12 +60,38 @@ public void write(DataOutput out) throws IOException {
...
+            LongWritable other = (LongWritable) o;
+            return this.value == other.get();
+        } else {
             return false;
-        IntWritable other = (IntWritable) o;
-        return this.value == other.value;
+        }
     }
 
     public int hashCode() {
```
AlexDBlack: If Int/Byte/LongWritable values can be equal, they should all use the same hashcode method (i.e., Long hashcode)

datavec-api/src/main/java/org/datavec/api/writable/LongWritable.java
```diff
@@ -58,16 +59,42 @@ public void write(DataOutput out) throws IOException {
...
+            IntWritable other = (IntWritable) o;
+            return this.value == other.get();
+        }
+        else { return false; }
     }
 
     public int hashCode() {
-        return (int) value;
+        // Defer to the long hashCode, which collides with integer hashCodes on relevant values
+        return Long.valueOf(value).hashCode();
```
AlexDBlack: Let's implement a version of Java 8's Long.hashCode method as a util method - no point producing extra garbage because we don't have a 1-line method...

huitseeker: Yes, of course.

datavec-api/src/main/java/org/datavec/api/writable/IntWritable.java
huitseeker: The original hashcodes of these in Java have this property already, and since we're reusing them, they will transfer to writable.

datavec-api/src/main/java/org/datavec/api/writable/ByteWritable.java
```diff
@@ -59,13 +60,31 @@ public void write(DataOutput out) throws IOException {
...
+        } else if (o instanceof  LongWritable) {
+            other = ((LongWritable) o).toDouble();
+        } else if (o instanceof ByteWritable) {
+            other = ((ByteWritable) o).toDouble();
+        } else if (o instanceof  DoubleWritable) {
+            other = ((DoubleWritable) o).toDouble();
+        } else if (o instanceof  FloatWritable) {
+            other = ((FloatWritable) o).toDouble();
+        } else { return false; }
+        return DoubleMath.fuzzyEquals(this.value, other, tolerance);
```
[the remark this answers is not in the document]
huitseeker: I would need this method to return a double even when the writable is not of a numeric type. Short of a crufty encoding (sometimes the numeric types will indeed want to use this double), properly writing this method will require unnecessary boxing. 

On a very secondary point, I find the case disjunction makes the cases we accept blatant.

datavec-api/src/main/java/org/datavec/api/writable/FloatWritable.java
```diff
@@ -59,16 +60,40 @@ public void write(DataOutput out) throws IOException {
...
             return false;
-        FloatWritable other = (FloatWritable) o;
-        return this.value == other.value;
+        }
     }
 
     public int hashCode() {
-        return Float.floatToIntBits(value);
+        // defer to Float.hashCode, which does what we mean for it to do
+        return Float.valueOf(value).hashCode();
```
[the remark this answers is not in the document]
huitseeker: Agreed on the boxing. As to float / double hashCode, the deference to Java makes this work if the Float/Double are equal-izable.

datavec-api/src/main/java/org/datavec/api/writable/FloatWritable.java
```diff
@@ -59,16 +60,40 @@ public void write(DataOutput out) throws IOException {
...
     public boolean equals(Object o) {
-        if (!(o instanceof FloatWritable))
+        if (o instanceof FloatWritable){
+            FloatWritable other = (FloatWritable) o;
+            return this.value == other.value;
+        }
+        if (o instanceof DoubleWritable){
+            DoubleWritable other = (DoubleWritable) o;
+            float otherFloat = (float) other.get();
+            return (other.get() == otherFloat && this.value == otherFloat);
```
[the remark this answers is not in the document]
huitseeker: I've added the case explicitly : they won't be equal. This downcasts the `Double` and returns true if and only if the downcast is w.l.o.g and to the comparison's float. I think it matches the instincts users will expect from how comparisons work in Java. If they want a more permissive match, `fuzzyEquals` is what should be used.

PR #412, opened 2017-08-31
NLP transforms now persist categorical metadata

turambar:
NLP transforms now persist categorical metadata after transforming to sequence of integers.

The NLP TextTo{Character,Integer}SequenceTransforms now set the minValue and maxValue properties of the destination Integer column when converting text to a sequence of indeces. This allows us to persist the vocabulary size.

## How was this patch tested?

Updated TestTransforms unit tests.


turambar:
Fixes this issue: https://github.com/deeplearning4j/DataVec/issues/413

PR #420, opened 2017-09-05
Allow leading zeros for NumberedFileInputSplit 

turambar:

With these changes, NumberedFileInputSplit permits two kinds of formatting -- "prefix4.suffix" or "prefix0004.suffix." Latter is more suitable for, e.g., Hadoop mapfiles spit out by a Spark ETL pipeline.

## How was this patch tested?

Added a NumberedFileInputSplitTests class with some unit tests to verify the name formatting works. I've also run it in live code, and it performed fine.

### DIFF
--- a/datavec-api/src/main/java/org/datavec/api/split/NumberedFileInputSplit.java
+++ b/datavec-api/src/main/java/org/datavec/api/split/NumberedFileInputSplit.java
@@ -1,19 +1,3 @@
-/*-
- *  * Copyright 2016 Skymind, Inc.
- *  *
- *  *    Licensed under the Apache License, Version 2.0 (the "License");
- *  *    you may not use this file except in compliance with the License.
- *  *    You may obtain a copy of the License at
- *  *
- *  *        http://www.apache.org/licenses/LICENSE-2.0
- *  *
- *  *    Unless required by applicable law or agreed to in writing, software
- *  *    distributed under the License is distributed on an "AS IS" BASIS,
- *  *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- *  *    See the License for the specific language governing permissions and
- *  *    limitations under the License.
- */
-
 package org.datavec.api.split;
 
 import org.datavec.api.util.files.UriFromPathIterator;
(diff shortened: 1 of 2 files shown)


turambar:
Addresses this issue: https://github.com/deeplearning4j/DataVec/issues/421

datavec-api/src/main/java/org/datavec/api/split/NumberedFileInputSplit.java
```diff
@@ -1,19 +1,3 @@
-/*-
```
huitseeker: This should not be removed

datavec-api/src/main/java/org/datavec/api/split/NumberedFileInputSplit.java
```diff
@@ -39,15 +25,18 @@
     private final int minIdx;
     private final int maxIdx;
 
+    private static final Pattern p = Pattern.compile("\\%(0\\d)?d");
```
huitseeker: To be compliant with `printf`-formatted strings, which is the format your users will expect to deal with, you have to at least document the following interactions in unit tests (and possibly do something better in the cases below)

```
scala> val pat = "\\%(0\\d)?d".r
pat: scala.util.matching.Regex = \%(0\d)?d

scala> def f(x: String): String = x match {
     |   case pat.unanchored(y) => s"Matched, with pattern $y"
     |   case _ => "UNMATCHED"
     | }
f: (x: String)String

scala> f("/path/to/files/prefix%5d.suffix")
res24: String = UNMATCHED

scala> f("/path/to/files/prefix%+5d.suffix")
res26: String = UNMATCHED

scala> f("/path/to/files/prefix%-5d.suffix")
res27: String = UNMATCHED

scala> f("/path/to/files/prefix%011d.suffix")
res28: String = UNMATCHED

scala> f("/path/to/files/prefix%101d.suffix")
res29: String = UNMATCHED

scala> f("/path/to/files/prefix%0505d.suffix")
res30: String = UNMATCHED
```

datavec-api/src/test/java/org/datavec/api/split/NumberedFileInputSplitTests.java
```diff
@@ -0,0 +1,67 @@
...
+    @Test
+    public void testNumberedFileInputSplitWithLeadingZeroesNoSuffix() {
+        String baseString = "/path/to/files/prefix-%d";
+        int minIdx = 0;
+        int maxIdx = 10;
+        runNumberedFileInputSplitTest(baseString, minIdx, maxIdx);
+    }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testNumberedFileInputSplitWithLeadingSpaces() {
```
huitseeker: O_o

PR #437, opened 2017-10-09
added asMatrix() for multipage and its test

BioinfoTongLI:

added the support of multipage tiff loading

## How was this patch tested?

tested with testasMatrix() in TestNativeImageLoader.java


datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
```diff
@@ -551,4 +553,27 @@ public INDArray asMatrix(ImageWritable writable) throws IOException {
...
+     */
+    public INDArray asMatrix(String path2tif) throws IOException {
+        PIXA pixa = pixaReadMultipageTiff(path2tif);
+        INDArray data = Nd4j.create(1, pixa.n(), pixa.pix(0).h(), pixa.pix(0).w());
+        INDArray currentD;
+        for (int i = 0; i < pixa.n(); i++) {
+            PIX pix = pixa.pix(i);
+            currentD = asMatrix(convert(pix));
+            pixDestroy(pix);
+            data.put(
```
BioinfoTongLI: in currentD, 
```
currentD.get(NDArrayIndex.all(),NDArrayIndex.point(1),NDArrayIndex.all())
```
is filled with 0

saudet:
LGTM, what do you think @AlexDBlack ?

BioinfoTongLI:
Hi, I think maybe the simplest solution for instance is to force users to send 8 bit images instead of 16 bit, which is not often used. So that we won't need to touch the convert method. But it depends on your needs, do you want to work on scientific images? If so, 16 bit support is often required, even 32bit ,64bit.

saudet:
It does not support 16-bit images yet, that's fine.

However, `asMatrix(String path)` is a confusing method name, and we don't need a new method anyway. This code should be part of where `pixReadMem()` is currently used. There is some code duplication there, so please create a new private method where to put all of that + the new code. I'm not sure how TIFF support works, but we could call `pixaReadMemMultipageTiff` by default on TIFF files and only when that fails fall back to `pixReadMem()`. What do you think?

BioinfoTongLI:
OK, so i saw 3 methods using `pixReadMem()`, the `asMatrixView`, the `asMatrix(InputStream is)` and the `asWritable(File f)`, we should in all these methods test whether the input is a tiff file. Then if it is, call  `pixaReadMemMultipageTiff`, If this fails, we call `pixReadMem()` is that what you want?

saudet:
Yes, and calling first ` pixaReadMemMultipageTiff`, then `pixReadMem` as fallback, is probably the best "TIFF detection" we're going to get...

BioinfoTongLI:
Yeah, the tiff detection can be tricky... Nevermind, i will look into that. BTW, maybe i should change the test image to an 8 bit image? so that i can for instance walkaround the the first "blank image"?

saudet:
Don't worry about the 16-bit image. We'll need to update convert() and test it out at some point anyway.

BioinfoTongLI:
@saudet i tried to catch exception from ` pixaReadMemMultipageTiff`, but it doesn't throw any. At least, at Java level. I do see a line of error from logger `Error in pixaReadMultipageTiff: file not tiff`,  but i think this is thorw from cpp that i can not catch. So i decided to test file extention of `File`, and it works fine. Your idea about `pixReadMem` is great, however ` pixaReadMemMultipageTiff` takes path to the tiff file as input, but not `byte[] bytes`. It would be difficult to put them in the same method, isn't it? What do you think?

BioinfoTongLI:
since you said `asMatrix(String path)` is confusing, i declared it as private. It doesn't change much at the code level. But one method less for other users, `INDArray asMatrix(File f)` will handle the format.

saudet:
Although `pixaReadMultipageTiff` reads from a file, `pixaReadMemMultipageTiff` most certainly does not take a filename as argument: It reads from a buffer in native memory. And returns an error code when it fails: It's a C function, not a C++ function.

BioinfoTongLI:
Ah, my bad! i used `pixaReadMultipageTiff`..... I will fix it

BioinfoTongLI:
bad news, `pixaReadMemMultipageTiff` is only avaliable in 1.7.4 of leptonica but not 1.7.3. Maybe you can publish a release of 1.7.4 for javacpp 1.3?

saudet:
Ah, I guess we'll have to update that...

saudet:
Ok, sorry it took a while, but Leptonica 1.74.4 is now in with the latest commit: https://github.com/deeplearning4j/DataVec/commit/36c4b9b9911b8c12645810f4ce37d1e4607babe6 !

Would you have time to finish this pull request?

BioinfoTongLI:
Great ! I'd love to, but maybe not immediately. I'll be more available in 3 weeks. If it's not urgent, I'll finish the rest of it.

saudet:
It's not urgent, no. Take you time! Thanks

BioinfoTongLI:
Hi, @saudet, I'm feeling that it's a little bit out of my capacity changing this simple part of loader... Please tell me if I did some naive errors.
 To cut long story short, I saw that you pushed your commits to the master. So I rebased this `datavec-0.9.1-multipage` thread onto the your current `master`.
But then I got a lot of `Not Found` errors in maven pom.xml. I believe it's because I dont't have the entire `deeplearning4j`?
Does this mean I have to build the entire `deeplearning4j` to modify the DataVec?
If I remember it well, you told me that in version 0.9.1 I dont have to do that.

I'm sorry this part has nothing to do with the `NativeImageLoader`, but without solving the problem I can not run the tests.

saudet:
You'll need to merge/rebase your fork on the master branch, yes. You don't need to compile all the other modules though; they are available as SNAPSHOT artifacts: https://deeplearning4j.org/snapshots

BioinfoTongLI:
I managed to run the `mvn test`. I got this error for all the other non-tiff images,

> tifferror: Not a TIFF or MDI file, bad magic number 55551 (0xd8ff).

 since you said that 

> Yes, and calling first pixaReadMemMultipageTiff, then pixReadMem as fallback, is probably the best "TIFF detection" we're going to get...

Is it coming from leptonica ? How can I catch that with Java ?
I tried `try...catch (Exception e)`, it seems to ignore my return in catch.
I'm barely know C++. Can you help me on that?

saudet:
`pixaReadMemMultipageTiff()` is probably just going to return `null` on error, so try to branch on that.

BioinfoTongLI:
Hi @saudet , the `pixaReadMemMultipageTiff` doesn't return null, it returns a `pixa` object with `n()` equals zero. So I used this.... It's kind of ugly but I have no other ideas. Maybe you got one ? BTW, there is an annoying thing. When I try to load any image with this `pixaReadMemMultipageTiff`, it sends error messages to the logger. As results, when I run tests, I see a lot of errors coming from this method. But they do pass. just annoying. One last thing is about the 16 bit images. It seems to be a lot of work for adapting 16bit image. I saw you used `pix.d()/8` to find the channel number. By chance, I have a 2-channel, 16-bit image and with the method you return a `16/8-channel` image. So it passes. But if I give it a 3-channel 16 bit. It will fail. Well, since it looks like the `convert()` is really an important method, I don't want to make it messy. I just added a TODO tag in the `asMatrix()` that I created.
Tell me what do you need me to do for this thread if there is any.

saudet:
Looks like the merge failed...  Could you try to update the master branch on your fork and try again?
https://help.github.com/articles/syncing-a-fork/

As for pixaReadMemMultipageTiff() not returning null or display error messages, that's fine. Try to load with OpenCV *before*, and if that fails it doesn't really matter what happens with Leptonica because operations that most people care about will succeed with OpenCV anyway. Thanks!

BioinfoTongLI:
Wired. I can not see why the merge failed. There were some conflicts but I resolved them already in my branch. Is it because I'm not pushing my master branch but the branch called `datavec-0.9.1-multipage`, I will be surprised if so... 
As for opencv, yes, it seems to me openCV is good at handling image formats now. Then we get back to the problem of `asMatrix()` with File Path. Because `imreadmulti()` takes String as input to open and returns a `MatVector` object. I think it's cleaner this way by testing the file extension to chose which opener to use. Is this what you want?

saudet:
No filename, we need to load from `InputStream` (memory). (Hey, `imreadmulti()` is nice! Too bad there is no `imdecodemulti()`... ) Both OpenCV and Leptonica are pretty good at loading from memory otherwise, so it should be fine.

saudet:
Hi, could you clean up your branch? It's still not merging properly: https://github.com/deeplearning4j/DataVec/pull/437/files

BioinfoTongLI:
Hi, I agree that there were quite some bizzar thing happening in the last commit.. But I'm sorry can't understand what is happening in the `datavec-jdbc` folder. 
I tried to pull the current remote master and make changes from that branch. But I have problems rebasing this `multipage 0.9.1 branch` on that.
So I propose two solutions:

1. I close this branch and create another PR from the master which is ready to be pushed.
2. I send you the three files that I modified. You will only need to copy-paste (they are from current master, so there shouldn't be conflicts)

Issue #347, opened 2017-07-06
Review BalancedPathFilter


AlexDBlack:
BalancedPathfilter: the idea is good - provide a more balanced distribution of classes for the training data.

In practice: it often ends up being:
(a) too aggressive, leaving only a very small training set behind, and
(b) unclear to the user exactly what the end result is (how many are left? how many were filtered?)

saudet:
Some ideas here:
https://github.com/deeplearning4j/DataVec/issues/222#issuecomment-297610572

AlexDBlack:
https://github.com/deeplearning4j/deeplearning4j/issues/5203

PR #439, opened 2017-10-14
Create ProportionalPathFilter.java

rangwank:

Added an experimental ratio-based sampling. This path filter is supposed to extend the functionality of the existing BalancedPathFilter by allowing certain labels to have more data than others. 

This is an EXPERIMENTAL work, no stability is guaranteed in any shape or form.

The proportionality is taken in as an integer array ratio. For example, the dataset contains label A and B where A is needed to have twice the data point compared to B. The ratio array would be {2,1}. The filter would take in the path, shuffle, group the path by labels, and interleave the labels as normally would in BalancedPathFilter.
In the case above, the labels would be interleaved in BalancedPathFilter as: AB AB AB AB ..., and in this filter as: AAB AAB AAB AAB ...

The minimum and maximum labels per path are also taken in as arrays. The change in one minimum or maximum for one label will proportionally affect other labels.
The number of data points for each label can be calculated by the following:

Maximum: minimum(r_i)*ratio_i; r_i = round(max_i/ratio_i)
Minimum: maximum(r_i)*ratio_i; r_i = round(min_i/ratio_i)

In the same example, supposed that the maximum of A is set to 50 and B is set to 30, this filter will only attempt to find 25 from B.

## How was this patch tested?

Initial release, no real testing is done in this version.

Please review

### DIFF
--- a/datavec-api/src/main/java/org/datavec/api/io/filters/ProportionalPathFilter.java
+++ b/datavec-api/src/main/java/org/datavec/api/io/filters/ProportionalPathFilter.java
@@ -1,0 +1,157 @@
+package org.datavec.api.io.filters;
+
+import org.datavec.api.io.labels.ParentPathLabelGenerator;
+import org.datavec.api.io.labels.PathLabelGenerator;
+import org.datavec.api.writable.Writable;
+
+import java.net.URI;
+import java.util.*;
+/**
+ * This is an EXPERIMENTAL work, no stability is guaranteed in any shape or form.
+ * This path filter is supposed to extend the functionality of the existing BalancedPathFilter by allowing certain labels to have more data than others. 
+ * 
+ * The proportionality is taken in as an integer array ratio. 
+ * For example, the dataset contains label A and B where A is needed to have twice the data point compared to B. The ratio array would be {2,1}
+ * The filter would take in the path, shuffle, group the path by labels, and interleave the labels as normally would in BalancedPathFilter.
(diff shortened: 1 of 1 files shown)


saudet:
Related to issue https://github.com/deeplearning4j/DataVec/issues/347

Sounds like a fine idea, but half of it is just copied BalancedPathFilter.
Could we instead reuse what we have there in some way?

/cc @eypros @AlexDBlack What do you guys think?

rangwank:
Normally, one would see the BalancedPathFilter would be the special case of ProportionalPathFilter where the ratio of all labels are equal. Reusing BalancedPathFilter here would be difficult, but the reverse should be relatively simple.

saudet:
So we could just update BalancedPathFilter with new optional parameters,
why not do that?

eypros:
I think that semantically one would more easily identify ProportionalPathFilter as a general case and BalancedPathFilter as the special case where all classes are equals. I guess that BalancedPathFilter could easily be expanded to include this case also but seeing it is quite possible to have to deal with quite unbalanced data it would be a bit bizarre to use it. SO this reason I would go with the separate ProportionalPathFilter.

saudet:
Sure, we could also rebase BalancedPathFilter on ProportionalPathFilter, that's fine.

AlexDBlack:
Closing this due to inactivity, and lack of tests. Feel free to reopen @saudet if necessary.

saudet:
I'd like it if we could rebase BalancePathFilter on this. @rangwank? @eypros?

PR #440, opened 2017-10-16
Add RecordMetaDataImageURI and NativeImageLoader.asMat()/asFrame()

saudet:

New functionality that is useful for object detection.

## How was this patch tested?

New unit tests pass and manually testing with example code also works as intended.

### DIFF
--- a/datavec-api/src/main/java/org/datavec/api/records/metadata/RecordMetaDataImageURI.java
+++ b/datavec-api/src/main/java/org/datavec/api/records/metadata/RecordMetaDataImageURI.java
@@ -1,0 +1,41 @@
+/*-
+ *  * Copyright 2016 Skymind, Inc.
+ *  *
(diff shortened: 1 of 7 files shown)


saudet:
We're just adding information to the metadata. The URI stays the same, so it should load the same. Am I missing something?

AlexDBlack:
Yeah, the labels though. If we don't override loadFromMetaData, then we end up calling BaseImageRecordReader.record(URI, DataInputStream), which will create an IntWritable label, instead of an NDArrayWritable label.

https://github.com/deeplearning4j/DataVec/blob/master/datavec-data/datavec-data-image/src/main/java/org/datavec/image/recordreader/BaseImageRecordReader.java#L404-L434

saudet:
Ok, it's done!

saudet:
Yes, we need it for the tests to pass with current ND4J...

PR #446, opened 2017-10-25
Add GeographicMidpointReduction

AlexDBlack:
cc @crockpotveggies

### DIFF
--- a/datavec-api/src/main/java/org/datavec/api/transform/transform/reduce/GeographicMidpointReduction.java
+++ b/datavec-api/src/main/java/org/datavec/api/transform/transform/reduce/GeographicMidpointReduction.java
@@ -1,0 +1,148 @@
+/*
+ *  * Copyright 2017 Skymind, Inc.
+ *  *
+ *  *    Licensed under the Apache License, Version 2.0 (the "License");
+ *  *    you may not use this file except in compliance with the License.
+ *  *    You may obtain a copy of the License at
+ *  *
+ *  *        http://www.apache.org/licenses/LICENSE-2.0
+ *  *
+ *  *    Unless required by applicable law or agreed to in writing, software
+ *  *    distributed under the License is distributed on an "AS IS" BASIS,
+ *  *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  *    See the License for the specific language governing permissions and
+ *  *    limitations under the License.
+ */
+
+package org.datavec.api.transform.transform.reduce;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.datavec.api.transform.metadata.ColumnMetaData;
+import org.datavec.api.transform.ops.IAggregableReduceOp;
+import org.datavec.api.transform.reduce.AggregableColumnReduction;
+import org.datavec.api.transform.schema.Schema;
+import org.datavec.api.writable.Text;
+import org.datavec.api.writable.Writable;
+
+import java.util.Collections;
+import java.util.List;
+
+public class GeographicMidpointReduction implements AggregableColumnReduction {
+
+    private String delim;
+    @Getter @Setter
+    private Schema inputSchema;
+
+    public GeographicMidpointReduction(String delim){
+        this.delim = delim;
+    }
+
+    @Override
+    public IAggregableReduceOp<Writable, List<Writable>> reduceOp() {
+        return new AverageCoordinateReduceOp(delim);
+    }
+
+    @Override
+    public List<String> getColumnsOutputName(String columnInputName) {
+        return null;
+    }
+
+    @Override
+    public List<ColumnMetaData> getColumnOutputMetaData(List<String> newColumnName, ColumnMetaData columnInputMeta) {
+        return null;
+    }
+
+    @Override
+    public Schema transform(Schema inputSchema) {
+        //No change
+        return inputSchema;
+    }
+
+    @Override
+    public String outputColumnName() {
+        return null;
+    }
+
+    @Override
+    public String[] outputColumnNames() {
+        return new String[0];
+    }
+
+    @Override
+    public String[] columnNames() {
+        return new String[0];
+    }
+
+    @Override
+    public String columnName() {
+        return null;
+    }
+
+    public static class AverageCoordinateReduceOp implements IAggregableReduceOp<Writable, List<Writable>> {
+        private static final double PI_180 = Math.PI / 180.0;
+
+        private String delim;
+
+        private double sumx;
+        private double sumy;
+        private double sumz;
+        private int count;
+
+        public AverageCoordinateReduceOp(String delim){
+            this.delim = delim;
+        }
+
+        @Override
+        public <W extends IAggregableReduceOp<Writable, List<Writable>>> void combine(W accu) {
+            if(accu instanceof AverageCoordinateReduceOp){
+                AverageCoordinateReduceOp r = (AverageCoordinateReduceOp)accu;
+                sumx += r.sumx;
+                sumy += r.sumy;
+                sumz += r.sumz;
+                count += r.count;
+            } else {
+                throw new IllegalStateException("Cannot combine type of class: " + accu.getClass());
+            }
+        }
+
+        @Override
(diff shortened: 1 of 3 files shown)


datavec-api/src/main/java/org/datavec/api/transform/transform/reduce/GeographicMidpointReduction.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @Override
+    public IAggregableReduceOp<Writable, List<Writable>> reduceOp() {
+        return new AverageCoordinateReduceOp(delim);
+    }
+
+    @Override
+    public List<String> getColumnsOutputName(String columnInputName) {
+        return null;
```
crockpotveggies: Do we not need to return a new column name? If null, does it just perform the op on the existing column? (Sorry not versed in the new API)

datavec-api/src/main/java/org/datavec/api/transform/transform/reduce/GeographicMidpointReduction.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @Override
+    public List<ColumnMetaData> getColumnOutputMetaData(List<String> newColumnName, ColumnMetaData columnInputMeta) {
+        return null;
```
crockpotveggies: Similar question as above.

datavec-api/src/test/java/org/datavec/api/transform/reduce/TestReductions.java
```diff
@@ -0,0 +1,92 @@
...
+import static org.junit.Assert.assertEquals;
+
+public class TestReductions {
+
+    @Test
+    public void testGeographicMidPointReduction(){
+
+        //http://www.geomidpoint.com/example.html
+        //That particular example is weighted - have 3x weight for t1, 2x weight for t2, 1x weight for t1
+        Text t1 = new Text("40.7143528,-74.0059731");
```
crockpotveggies: Maybe move the coordinates to a polar extreme? That will exaggerate any mathematical issues.

datavec-api/src/main/java/org/datavec/api/transform/transform/reduce/GeographicMidpointReduction.java
crockpotveggies: Yea so in testing I'm getting NPEs from the reducer when this is null
```
...java.lang.NullPointerException
  at java.util.ArrayList.addAll(ArrayList.java:577)
  at org.datavec.api.transform.reduce.Reducer.transform(Reducer.java:130)
  at org.datavec.api.transform.sequence.window.ReduceSequenceByWindowTransform.transform(ReduceSequenceByWindowTransform.java:65)
```

crockpotveggies: The code that triggers the NPE
```
val transform = new TransformProcess.Builder(schema)
    .removeAllColumnsExceptFor("Timestamp","MMSI","Lat","Lon")
    .transform(new StringToTimeTransform("Timestamp","dd/MM/YYYY HH:mm:ss",DateTimeZone.UTC))
    .transform(new ConcatenateStringColumns("LatLon", ",", List("Lat","Lon")))
    .convertToSequence("MMSI", new NumericalColumnComparator("Timestamp", true))
    .transform(
        new ReduceSequenceByWindowTransform(
            new Reducer.Builder(ReduceOp.Count).keyColumns("MMSI")
                .countColumns("Timestamp")
                .customReduction("LatLon", new Reductions.GeoAveragingReduction)
                .takeFirstColumns("Timestamp")
                .build(),
            new TimeWindowFunction("Timestamp",10L,TimeUnit.MINUTES)
        )
    )
    .removeAllColumnsExceptFor("LatLon")
    .build
```

datavec-api/src/main/java/org/datavec/api/transform/transform/reduce/GeographicMidpointReduction.java
```diff
@@ -0,0 +1,158 @@
...
+            double z = Math.sin(lat);
+
+            sumx += x;
+            sumy += y;
+            sumz += z;
+            count++;
+        }
+
+        @Override
+        public List<Writable> get() {
```
huitseeker: This one and the above LGTM.

datavec-api/src/test/java/org/datavec/api/transform/reduce/TestReductions.java
huitseeker: Actually, you may want to test the ka-boom : two polar opposites, or more generally equally-distributed points on the sphere. Their Ka-boom nature should be apparent (i.e. raise an exception).

datavec-api/src/main/java/org/datavec/api/transform/transform/reduce/GeographicMidpointReduction.java
```diff
@@ -0,0 +1,158 @@
...
+            String[] split = str.split(delim);
+            if(split.length != 2){
+                throw new IllegalStateException("Could not parse lat/long string: \"" + str + "\"" );
+            }
+            double latDeg = Double.parseDouble(split[0]);
+            double longDeg = Double.parseDouble(split[1]);
+
+            double lat = latDeg * PI_180;
+            double lng = longDeg * PI_180;
+
```
huitseeker: The following is nice but costly, I'd like to have an idea of how to bail if we were to deal with infinitesimally close points.

datavec-api/src/main/java/org/datavec/api/transform/transform/reduce/GeographicMidpointReduction.java
```diff
@@ -0,0 +1,158 @@
+/*
```
huitseeker: This file seems to not be in the right place. I believe you want

datavec-api/src/main/java/org/datavec/api/transform/transform/reduce/GeographicMidpointReduction.java
huitseeker:
```

AlexDBlack: Fixed

datavec-api/src/main/java/org/datavec/api/transform/transform/reduce/GeographicMidpointReduction.java
AlexDBlack: Fixed

datavec-api/src/test/java/org/datavec/api/transform/reduce/TestReductions.java
AlexDBlack: Added exception, based on "special case" / pt 9 here:
http://www.geomidpoint.com/calculation.html

datavec-api/src/main/java/org/datavec/api/transform/transform/reduce/GeographicMidpointReduction.java
AlexDBlack: Maybe overkill? But if one of you wants to implement rounding error reduction, by all means go ahead. As to cache of last point: hm, that implies order, which I don't think we can assume in general (like, spark execution, or even just non-sequences).

datavec-api/src/main/java/org/datavec/api/transform/transform/reduce/GeographicMidpointReduction.java
AlexDBlack: Moved to org.datavec.api.transform.reduce.impl.GeographicMidpointReduction

datavec-api/src/main/java/org/datavec/api/transform/transform/reduce/GeographicMidpointReduction.java
huitseeker: Yeah, except I *know* @crockpotveggies is going to use this on a list where 50% of points or more are within meters of each other. I guess he'll have to implement the above then :)

PR #449, opened 2017-10-31
Add SvhnLabelProvider

saudet:

A label provider for datasets at http://ufldl.stanford.edu/housenumbers/.

## How was this patch tested?

Currently only manually tested. I could make a unit test, but we'd basically need the original data file (~100 MB) or create a small one with MATLAB, which I don't have access to these days...


saudet:
Unfortunately Octave doesn't support the HDF5 variant: http://wiki.octave.org/FAQ#Mat-File_format :(

Yeah, probably not worth the effort. It's only to get a neat example running for now.

PR #469, opened 2017-11-29
RecordReader.resetSupported() and concatenating record reader

AlexDBlack:
Adds RecordReader.resetSupported(), and ConcatenatingRecordReader

See also:
https://github.com/deeplearning4j/deeplearning4j/pull/4345
https://github.com/deeplearning4j/deeplearning4j/issues/4339

### DIFF
--- a/datavec-api/src/main/java/org/datavec/api/split/InputStreamInputSplit.java
+++ b/datavec-api/src/main/java/org/datavec/api/split/InputStreamInputSplit.java
@@ -43,8 +43,7 @@ public class InputStreamInputSplit implements InputSplit {
      * @param path the path to use
      */
     public InputStreamInputSplit(InputStream is, String path) {
-        this.is = is;
-        this.location = new URI[] {URI.create(path)};
+        this(is, URI.create(path));
     }
 
     /**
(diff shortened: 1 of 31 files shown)


datavec-api/src/main/java/org/datavec/api/split/InputStreamInputSplit.java
```diff
@@ -97,6 +97,19 @@ public long length() {
...
+            is.reset();
+        } catch (IOException e){
+            throw new RuntimeException("Error reseting input stream",e);
+        }
+    }
+
+    @Override
+    public boolean resetSupported() {
+        if( is != null){
+            return is.markSupported();
```
saudet: When markSupported() == true, wouldn't we need to call mark() somewhere at the beginning for this to be valid?

AlexDBlack: Good question, and you may be right. I should probably add a test for that...

AlexDBlack:
I have decided to disallow resetting on all streams... re-reading the javadoc for mark: https://docs.oracle.com/javase/7/docs/api/java/io/InputStream.html#mark(int)

> The general contract of mark is that, if the method markSupported returns true, the stream somehow remembers all the bytes read after the call to mark and stands ready to supply those same bytes again if and whenever the method reset is called. However, the stream is not required to remember any data at all if more than readlimit bytes are read from the stream before reset is called.

In other words, we have to specify a read limit value... In some cases, this might result in an internal storage buffer being used.
If the readLimit value is too small, reset isn't possible. If it's too large, we could use an unreasonable amount of memory (depending on the underlying stream implementation). Consequently, there's no single readLimit value that will work in all cases...

saudet:
Sounds good to me!

PR #479, opened 2017-12-21
Add .travis.yml and build script for Linux

saudet:

Let Travis CI make builds and deploy SNAPSHOT artifacts to Sonatype OSS automatically.

## How was this patch tested?

Builds pass on Travis CI and SNAPSHOT artifacts are deployed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+branches:
+  only:
+    - master
 dist: trusty
 sudo: false
 cache:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,26 +1,23 @@
-language: java
-
+dist: trusty
```
agibsonccc: Still not sure on distro here, it seems like we're mainly assuming rpms in our builds.

saudet: No choice, only thing supported by Travis. To do RPMs, we can use Docker in the relevant project for that.

Issue #502, opened 2018-02-06
(Presumably) incorrect input to FFT in Spectrogram/FastFourierTransform


DrChainsaw:
#### Issue Description

Spectrogram through FastFourierTransform gives com.sun.media.sound.FFT.transform a double array of real valued signal amplitudes while it expects a complex double array with same format as the output (even indexes are real and odd are imaginary). 

The result is that the output can to the extent of my knowledge not really be called the fft of the signal. It also has the more visible consequence that the fft produced is only a quarter of the signal length instead of half (half is expected since negative freqs are discarded).

I don't know if this is some kind of Deep Learning trick, but at least for me the accuracy of my classifier improved slightly when I changed the implementation.

Something like this does the trick (sorry about lack of formatting):

 final int sampleSize = 2*signals[0].length;
 FFT fft = new FFT(sampleSize/2, -1);

double[] signalsAsComplex = new double[sampleSize];
for(int j = 0; j < sampleSize; j+=2) {
      signalsAsComplex[j] = signals[i][j/2];
      signalsAsComplex[j+1] = 0; // Can be omitted if one trusts in initialization
}
fft.transform(signalsAsComplex); 

Obviously if one wants to retain the possibility to fft complex signals the code above needs to go into Spectrogram.java instead.

#### Version Information

Present in latest version of Spectrogram (a8744ac on Jul 10 2017) /FastFourierTransform (2850231 Feb 28, 2017)

#### Contributing

Trivial fix. Unsure about desired approach wrt complex signals and model backwards compatibility (e.g. third parties models build on this version of the file).

saudet:
I think the current implementation is just not correct, so please make the necessary modifications and send a pull request! BTW, we should probably switch to FFTW for performance reasons and because  com.sun.media.sound.FFT is not available on Android or with Java 9 anymore:
https://github.com/bytedeco/javacpp-presets/tree/master/fftw
However, since that is under the GPL license, we'd have to disable it by default as with FFmpeg:
https://github.com/deeplearning4j/DataVec/blob/master/datavec-data/datavec-data-codec/pom.xml#L53

PR #505, opened 2018-02-07
Spectrogram fft fix

DrChainsaw:

Fixes for issue #502 

Length of each timewindow in Spectrogram now is same sa fftSampleSize.
FastFourierTransform has a method which takes a real-valued signal. I left the old behaviour (which assumed complex valued) since there was a testcase for it already. 

## How was this patch tested?

New testcase for FFT added, although it has magic numbers in it. Manual inspection in debugger to see that change of real to complex format expected by FFT is correct.
Spectrogram windowing was tested manually by (temporarily) adding amplitudes as argument to buildSpectrogram, commenting out calling it from constructor and a main method like this:
    public static void main(String[] args) {
        Spectrogram spec = new Spectrogram(null, 4,2);
        short[] testVec = {0,1,2,3,4,5,6,7};
        spec.buildSpectrogram(testVec);
    } 

Please review

### DIFF
--- a/datavec-data/datavec-data-audio/src/main/java/org/datavec/audio/dsp/FastFourierTransform.java
+++ b/datavec-data/datavec-data-audio/src/main/java/org/datavec/audio/dsp/FastFourierTransform.java
@@ -28,10 +28,16 @@ public class FastFourierTransform {
     /**
      * Get the frequency intensities
      *
-     * @param amplitudes real-valued amplitudes of the signal
+     * @param amplitudes amplitudes of the signal. Format depends on value of complex
+     * @param complex if true, amplitudes is assumed to be complex interlaced (re = even, im = odd), if false amplitudes
+     *                are assumed to be real valued.
      * @return intensities of each frequency unit: mag[frequency_unit]=intensity
      */
-    public double[] getMagnitudesForReal(double[] amplitudes) {
+    public double[] getMagnitudes(double[] amplitudes, boolean complex) {
+
+        if(complex) {
+            return getMagnitudes(amplitudes);
+        }
 
         // FFT expects complex input where even indexes are real parts
         // and odd indexes are img parts.
@@ -43,7 +49,7 @@ public class FastFourierTransform {
             amplitudesAsComplex[j] = amplitudes[j / 2];
             amplitudesAsComplex[j + 1] = 0;
         }
-        return getMagnitudesForComplex(amplitudesAsComplex);
+        return getMagnitudes(amplitudesAsComplex);
     }
 
     /**
@@ -52,7 +58,7 @@ public class FastFourierTransform {
      * @param amplitudes complex-valued signal to transform. Even indexes are real and odd indexes are img
      * @return intensities of each frequency unit: mag[frequency_unit]=intensity
      */
-    public double[] getMagnitudesForComplex(double[] amplitudes) {
+    public double[] getMagnitudes(double[] amplitudes) {
 
         int sampleSize = amplitudes.length;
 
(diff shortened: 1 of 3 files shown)


datavec-data/datavec-data-audio/src/main/java/org/datavec/audio/dsp/FastFourierTransform.java
```diff
@@ -20,20 +20,39 @@
...
     /**
      * Get the frequency intensities
-     * 
-     * @param amplitudes
-     *            amplitudes of the signal
+     *
+     * @param amplitudes real-valued amplitudes of the signal
      * @return intensities of each frequency unit: mag[frequency_unit]=intensity
      */
-    public double[] getMagnitudes(double[] amplitudes) {
```
saudet: @DrChainsaw We could keep backward compatibility by adding the following lines, what do you think?
```java
/** Returns {@code getMagnitudes(amplitudes, true)}. */
public double[] getMagnitudes(double[] amplitudes) {
    return getMagnitudes(amplitudes, true);
}
/** Returns {@code complex ? getMagnitudesForComplex(amplitudes) : getMagnitudesForReal(amplitudes)}. */
public double[] getMagnitudes(double[] amplitudes, boolean complex) {
    return complex ? getMagnitudesForComplex(amplitudes) : getMagnitudesForReal(amplitudes);
}
```

DrChainsaw: Well, its your call. 

I was thinking that this might be a rare case when backwards compatibility is not desired. Chances are that people have been using the method without realizing it assumes the wierd odd-even complex format from FFT.transform and maintaining backwards comptibility would keep masking that potential issue. Otoh, who am I to second guess what people do and do not think about and break their code because of it? :)

I have also become somewhat allergic to methods with flag arguments over the years, but I'll gladly make the change if you prefer it. If it is just backwards compatibility you were after that obviously can be achieved by just calling getMagnitudesForComplex right away. If you prefer the flag way I assume you'd also like ForReal and ForComplex be private, right?

saudet: No, the flag is just to indicate that it has a "default value" via the other overload. Makes it harder for people to miss it IMO...

DrChainsaw:
@saudet Sorry for the delay. Noob as I am I accidentally pushed changes to the fork master and not to the branch.

I'm not sure this is exactly what you were after, but I assumed you didn't want four public methods to do the two things the class can do.

saudet:
Well, looks like a good enough compromise for me for now. 👍  Thanks!

We'll need to revist this anyway to get rid of the com.sun.media.sound.FFT class at some point since it's not available in Java 9 anymore.

DrChainsaw:
@saudet Glad I could help!

Just a tip regarding the FFT: I recently converted my project to use Jtransform for transforms. It's licensed under BSD-2 Clause which from what I understand is a nicer license compared to GPL as it is non-viral. They claim comparable performance to FFTW (although the min execution time benchmark methodology looks somewhat fishy to me). Having a mix java and c in your project is bound to fail anyways /jk :)

I haven't done any measurements, but to me, there is no perceptable difference for ETL when training on raw waveforms vs spectrograms/MFSC/MFCC etc. It seems to be dominated by the file reading. For "online" inference (classifying audio stream from soundcard) even with my worst model ensembles the program spends ~70% of the time idling while the sound buffer fills up. All in all, FFT performance does not seem to be that important since other things (GPU memory for me) is bottlenecking. 

I also saw that NDArrays have a built in fft method. I haven't tried it yet though but I guess it could be a kind of nice self-containment thing.

saudet:
https://github.com/wendykierp/JTransforms looks awesome! If you could integrate that it would be great! FFT isn't currently implemented in ND4J and since this is currently only for ETL, I think it makes sense to just use JTransforms as is in DataVec. But if you'd like to create the API in ND4J for FFT first, and then use that in DataVec, that would be fine too.

PR #508, opened 2018-02-15
Change of FFT implementation to Jtransforms

DrChainsaw:

Changed FFT implementation to JTransforms as discussed in #505.
I might have gone a bit overboard with the backwards compatibility this time around: Previous implementation returned an array which was four times smaller than the input when input was complex. 
This is what you get with the method without the boolean argument while the one with the boolean arg returns an array of half the size as the input for both complex and real amplitudes. 

## How was this patch tested?

Unit tests

Please review

### DIFF
--- a/datavec-data/datavec-data-audio/src/main/java/org/datavec/audio/dsp/FastFourierTransform.java
+++ b/datavec-data/datavec-data-audio/src/main/java/org/datavec/audio/dsp/FastFourierTransform.java
@@ -16,7 +16,7 @@
 
 package org.datavec.audio.dsp;
 
-import com.sun.media.sound.FFT;
+import org.jtransforms.fft.DoubleFFT_1D;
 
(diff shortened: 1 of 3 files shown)


agibsonccc:
@DrChainsaw this is great, I'll let @AlexDBlack comment on this but I don't have any objections so far.

AlexDBlack:
I scanned this quickly, looks fine to me.
I'll defer to @saudet for the final review/merge.

treo:
To me it looks like there should be a slight change: As this only pertains to datavec-data-audio, the jtransforms dependency should be placed in the corresponding pom.xml. Putting it right into the main pom.xml makes all datavec packages depend on it.

Ngosti2000:
@treo ..yeap makes sense to separate as my android build is failing..have raised the issue on it as well. #4718

PR #511, opened 2018-02-23
CIFAR loader workspace fixes

AlexDBlack:
***WIP DO NOT MERGE***

Attempts to fix: https://github.com/deeplearning4j/deeplearning4j/issues/4673
Seems to still occur, however. Basic design of CIFAR loader is problematic - it loads all data into memory, which isn't a good idea in general.


AlexDBlack:
Closing this for now instead of leaving it hanging - might revisit at a later date.

PR #515, opened 2018-02-27
Move Jtransforms dependency to pom.xml of datavec-data-audio

DrChainsaw:
Move Jtransforms dependency to pom.xml of datavec-data-audio as requested by @treo in #508. I kept the version property in the main pom.xml as this seemed to be the way it is done for other dependencies.

I don't know why this pull request shows changes which are already merged. A quick search lead me to some SO-thread which stated that this is some limitation in github, but I might just as well have done something wrong. Only "real" changes are in the poms.

## How was this patch tested?

Unit tests in datavec-data-audio

Please review

### DIFF
--- a/datavec-api/pom.xml
+++ b/datavec-api/pom.xml
@@ -156,22 +156,4 @@
      </dependency>
     </dependencies>
 
-    <profiles>
-        <profile>
-            <id>backport-fun-java8</id>
-            <activation>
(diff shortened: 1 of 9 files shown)


DrChainsaw:
Ok, I just realized this was already fixed. Sorry for the inconvenience.

PR #518, opened 2018-03-11
add asMatrix() to load multipage tiff

BioinfoTongLI:

a private asMatrix() method to load buffered multipage tiff using Leptonica

## How was this patch tested?

test added in original `testAsMatrix()`

Please review


BioinfoTongLI:
Hi, @saudet. I tried what you said. The problem is that imdecode() doesn't fail... It does read the multipage tiff, but only the first slice. So neither `null` or `empty` test will fire. The solution is to know beforehand the number of slice is large than 1 or not for current input stream which I think can be tricky.

saudet:
I see... We could patch Leptonica to disable error messages, like it already does for warning messages:
https://github.com/DanBloomberg/leptonica/commit/d42a51d29eb8bd086ddee83dd12bab95a9b68238

saudet:
Ok, here's what we're going to do. Let's add overloads `asMatrix(File f, boolean multipage)`, `asMatrix(InputStream is, boolean multipage)` with default value of `false` so that it doesn't try to use Leptonica when it doesn't need to. What do you think?

BioinfoTongLI:
Sorry, I was busy lately. I'm a little bit free this week. I will look into that.

BioinfoTongLI:
Finally, I'm back. Sorry for the lagging. So, I believe you want me to write another constructor of `NativeImageLoader`. But there is already a constructor with `NativeImageLoader(int height, int width, int channels, boolean centerCropIfNeeded)`. Do you want an append of this one, like `NativeImageLoader(int height, int width, int channels, boolean centerCropIfNeeded, boolean isMultipage)` or do you have better ideas?

saudet:
I'd see an enum MultiPageMode or something actually, with values like MINIBATCH and CHANNELS. What do you think?

BioinfoTongLI:
There is one thing I don't understand, what is the difference between MINIBATCH and CHANNELS? It's just another dimension beyond X,Y, no?

If so, I believe the fact that you want to use enum is because we may want to load more complex data (like in Microscopic data with 5 dimensions, x, y, z, time, position). But do you think user will load images with that complexity (except me, of course...) ? From my personal point of view, you may just want to know whether or not it's a multipage file. Then it's up to user to sort it once they got the multi-dimension matrix, no?

Since I don't know what is the direction of deeplearning4j, I can not say which one is better. But if you want apply Datavec one scientific project, then yes. We should use enum and switch it to handle different formats, otherwise I'd say a true/false is OK for now.

BioinfoTongLI:
Sorry, maybe it's obvious for you, but I don't understand what is the difference between minibatch and channels.
An image is [x,y]. From my experience, a list like [[x1,y1,], [x2,y2], [x3,y3]] is a 3-channel images. And it can also be a minibatch of 3 images also, no? Do you mean [[x,y]] for `minibatch`?
As for the `int` or `enum` problem. I'd say maybe just go for `enum`.  It's more intuitive with `String`.
Then, where will this `enum` be ? a standalone public ? in-class Protected ? not private I believe. I'd say public in an isolated `MultiPageMode.java`. Is it OK?

saudet:
Ah, I'm sorry, I thought you already understood because you wrote this line:
```
new INDArrayIndex[]{NDArrayIndex.all(), NDArrayIndex.point(i),NDArrayIndex.all(),NDArrayIndex.all()},
```
This maps each image to a channel. Some users might want to map them to different images in the same batch, like this:
```
new INDArrayIndex[]{NDArrayIndex.point(i),NDArrayIndex.all(),NDArrayIndex.all(),NDArrayIndex.all()},
```
Since we should provide that choice, and it's only a single parameter, for this single line of code (along with `Nd4j.create(1, pixa.n(), pixa.pix(0).h(), pixa.pix(0).w())` -> `Nd4j.create(pixa.n(), channels, pixa.pix(0).h(), pixa.pix(0).w())` actually), I thought we might as well do it together with the choice to disable multipage entirely, that's all.

If you create an enum, put it either as a `public static` in the class, or as `public` in the same package. Either are fine. Thanks!

BioinfoTongLI:
Ah, OK. I almost forgot this is for machine learning... Now, it's clear. Thanks.

BioinfoTongLI:
I'm blocked with unittest in datavec-api...
I've never changed this api part. So maybe someone else has changed this part without unittesting ? That's wired. I didn't have this before.

> Tests in error: 
>   CSVSequenceRecordReaderTest.testCsvSeqAndNumberedFileSplit:239 » MissingFormatArgument
> 
> Tests run: 250, Failures: 0, Errors: 1, Skipped: 1
> 
> [INFO] ------------------------------------------------------------------------
> [INFO] Reactor Summary:
> [INFO] 
> [INFO] DataVec ............................................ SUCCESS [  1.612 s]
> [INFO] datavec-api ........................................ FAILURE [ 13.178 s]

I tried to skip all the other unittests and run only the ones in `TestNativeImageLoader` with
`mvn -Dtest=TestNativeImageLoader test`
and 
`mvn test -DfailIfNoTests=false -Dtest=TestNativeImageLoader -pl datavec-data-image` for multi-module project. None is working.
This was working fine before. But not any more now. Do you know how to solve this ?
Or maybe you can just have a go on  `mvn test` to see whether it works?
I've tested on my mac and desktop (Ubuntu). They have both the same error above

AlexDBlack:
I can confirm that all DataVec tests are currently passing on master, and were also passing when I last tested (14 hours ago?). Before that (24-48  hours ago) there may have been some failures.
I'm guessing that you might simply need to rebase your PR to current DataVec master and try again.

BioinfoTongLI:
Thanks for both ! Now it's working as what I want.

datavec-data/datavec-data-image/src/test/java/org/datavec/image/loader/TestNativeImageLoader.java
```diff
@@ -98,20 +98,51 @@ public void testConvertPix() {
...
         assertEquals(4, array6.rank());
-        assertEquals(1, array6.size(0));
-        assertEquals(pages, array6.size(1));
+        assertEquals(pages, array6.size(0));
+        assertEquals(ch4, array6.size(1));
         assertEquals(h4, array6.size(2));
         assertEquals(w4, array6.size(3));
+
+        int ch5 = 4, pages1 = 1;
+        NativeImageLoader loader6 = new NativeImageLoader(h4, w4, 1, NativeImageLoader.MultiPageMode.CHANNELS);
```
BioinfoTongLI: I have to initialize the loader with channel number equal 1. Yet, actually I receive an INDArray with channel number != 1. 
Here is the problem. Since if I declare the channel number != 1, this line `asMatrix(convert(pix))` will give me an INDArray with shape (1, channel #, h, w).
I think this is because in `transformImage`, it is using the global variable "channels" during conversion. So by default, an image with the final dimension (1, channel #, h, w)  is created while I just need INDArray with (1,1,h,w).

BioinfoTongLI:
I got something "working" when I cheat a little. I pointed the issue out in the comment. Sorry if it's not clear.
To cut long story short, I think we need to change some fundamental method(s), like `transformImage` in the Loader. 
I would like to commit what I did before changing those parts.

BioinfoTongLI:
Hi, it seems that all tests have passed without any changes... Maybe I was wrongly testing. Now you can take a look at this to see if there is any modifications to do.

datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
```diff
@@ -683,4 +704,56 @@ public Mat asMat(INDArray array, int dataType) {
...
+        INDArray currentD;
+        INDArrayIndex[] index = null;
+        switch (this.multiPageMode) {
+            case MINIBATCH:
+                data = Nd4j.create(pixa.n(), 1, pixa.pix(0).h(), pixa.pix(0).w());
+                break;
+            case CHANNELS:
+                data = Nd4j.create(1, pixa.n(), pixa.pix(0).h(), pixa.pix(0).w());
+                break;
+            default:
```
saudet: Let's use `FIRST` explicitly here, and have `default` throw an `UnsupportedOperationException`.

datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
```diff
@@ -683,4 +704,56 @@ public Mat asMat(INDArray array, int dataType) {
...
+                case CHANNELS:
+                    index = new INDArrayIndex[]{NDArrayIndex.all(), NDArrayIndex.point(i),NDArrayIndex.all(),NDArrayIndex.all()};
+                    break;
+            }
+            data.put(index , currentD.get(NDArrayIndex.all(), NDArrayIndex.all(),NDArrayIndex.all()));
+        }
+
+        return data;
+    }
+    public enum MultiPageMode{
```
saudet: We should probably move the enum to the beginning of the class before the constructors for clarity...

Issue #555, opened 2018-04-09
BaseImageRecordReader initialization with configuration overrides appendLabel


HDouss:
#### Issue Description

When initializing an `ImageRecordReader` with a passed configuration, the initialization overrides `appendLabel` to false. So, if the `ImageRecordReader` is instantiated with a label generator, and then, initialized with a configuration, many issues arise because `appendLabel` was set to false

I expect the initialization to keep the existing value of `appendLabel` if not set in the configuration like it is the case for the height and the width.
#### Version Information

* Deeplearning4j version : 0.9.1

AlexDBlack:
https://github.com/deeplearning4j/DataVec/pull/554

PR #556, opened 2018-04-09
#555 value for appendLabel is the already defined one by default

HDouss:

Fix for #555. Value for appendLabel is the already defined one by default. So if the configuration does not define a value the already defined value (at instantiation for example).
I think the correction is too simple for a test case.


AlexDBlack:
Thanks - I'll close this without merging, looks like the fix for #555 was already included in your other PR here: https://github.com/deeplearning4j/DataVec/pull/554

Issue #558, opened 2018-04-11
Inconsistent RecordReader listeners invocation


HDouss:
While trying to execute the `MnistImagePipelineExample`, I encountered some inconsistent behavior. The first time a record is read `DataSetIterator` delegates to `ImageRecordReader` to load a first record and then builds a `RecordReaderMultiDataSetIterator` that will be used for further readings. During that build, the `ImageRecordReader` is reset.

The problem appears when attaching a listener to the `ImageRecordReader`. The first time a record is read, the listeners are invoked, but this call is somehow useless because the `ImageRecordReader` is reset anyway. On further "batch" record readings, the listeners are not invoked at all.

Expected behavior is that the listeners are invoked when a record is actually and usefully read.
Encountered behavior is that:
* the listeners are invoked during the first record read that was not an actual data read but used only to build RecordReaderMultiDataSetIterator (`BaseImageRecordReader.next` method)
* the listeners are never invoked for further records "batch" readings (`BaseImageRecordReader.next(int num`) method)

Deeplearning4j version : 0.9.1

I will probably make a PR with a failing unit test in the upcoming days.

saudet:
Thanks! Will be waiting.

HDouss:
It could take sometime. Meantime it is very easy to reproduce, just run the `MnistImagePipelineExample` and notice that the `LogRecordListener` is invoked only once, and for an image that won't be the first to read (because reader will be reset afterwards). I expect: 
* the listener to be called for every record read
* the listener to be called only for the records actually read, i.e : not for a record that is read (to build RecordReaderMultiDataSetIterator) and then be "thrown" away

saudet:
Fixed with pull #564. Thanks for reporting and testing!!

HDouss:
Please note that the behavior below is not fixed:
* the listener to be called only for the records actually read, i.e : not for a record that is read (to build RecordReaderMultiDataSetIterator) and then be "thrown" away

But this is probably a bug in deeplearning4j. Notice the [initializeUnderlying](https://github.com/deeplearning4j/deeplearning4j/blob/master/deeplearning4j-data/deeplearning4j-datavec-iterators/src/main/java/org/deeplearning4j/datasets/datavec/RecordReaderDataSetIterator.java#L218) method that reads a record (and thus invokes the listener) for not an actually used record.

AlexDBlack:
@HDouss mind opening an issue on DL4J?
Not an priority IMO but we should definitely clean that up at some point.

PR #564, opened 2018-04-13
558

HDouss:

Add a unit test for inconsistent behavior for #558 
2 unit tests are added, the one showing the inconsistent behavior is annotated with @Ignore


agibsonccc:
@saudet can we merge this?

saudet:
I guess we could, but we still need to actually implement something for the failing test.

Issue #571, opened 2018-04-27
Reuse buffers in BaseImageRecordReader


AlexDBlack:
The byte[] allocations here can add up to a lot of memory and GC thrashing:

https://github.com/deeplearning4j/DataVec/blob/9a104e7cd8d759e5fc7f3eac1bd01f5c5454bff0/datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java#L404

![image](https://user-images.githubusercontent.com/2360237/39344754-b2361452-4a27-11e8-8577-b3aabbb73481.png)

AlexDBlack:
Some thoughts here:
A simple solution is keep the largest (so far) required byte[] buffer in NativeImageLoader... it would have to be per thread, or a simple pool (shared between threads). try/catch on loading, allocate a larger one if required.

The only issue with that is that what happens if we're done with image loading in our program? that byte[] can never be cleanup up by the GC. But given it's the raw File bytes (I think?) it should only be a few MB, maybe this doesn't matter.

A slightly more sophisticated option might be a pool/queue of weak references - reuse the byte[] if available, but allow it to be GC. Reliable and easy - but the downside of that is that we end up with more allocations/deallocations than if we just kept a hard reference.

Finally (perhaps a bit ugly) we could add a scheduled task to the queue - clear the reference if it doesn't get used for say 30 second.

saudet:
NativeImageLoader already maintains a state and it's not designed to be used across threads, so we can keep a buffer around in a field without any issue. It'll get garbage collected when that instance of NativeImageLoader also gets garbage collected.

AlexDBlack:
@saudet Ah, good point. The byte[] doesn't need to be static. OK, that makes it way easier. :)

PR #573, opened 2018-04-30
#571 NativeImageLoader buffer reuse

AlexDBlack:
***WIP DO NOT MERGE***
Fixes: https://github.com/deeplearning4j/DataVec/issues/571

All tests pass, good to go except for loading issue noted below (which should be resolved first).

@saudet without adding an Nd4j.create(1) at the top of the test (and 1 or 2 others now) I'm getting the following (when the tests are run in isolation):
```
java.lang.RuntimeException: No native JavaCPP library in memory. (Has Loader.load() been called?)
	at org.bytedeco.javacpp.BytePointer.<init>(BytePointer.java:103)
	at org.datavec.image.loader.NativeImageLoader.asMatrix(NativeImageLoader.java:211)
	at org.datavec.image.loader.NativeImageLoader.asMatrix(NativeImageLoader.java:202)
	at org.datavec.image.loader.TestNativeImageLoader.testBufferRealloc(TestNativeImageLoader.java:330)
...
Caused by: java.lang.UnsatisfiedLinkError: org.bytedeco.javacpp.BytePointer.allocateArray(J)V
	at org.bytedeco.javacpp.BytePointer.allocateArray(Native Method)
	at org.bytedeco.javacpp.BytePointer.<init>(BytePointer.java:98)
	... 25 more
```

What's the best workaround/solution here?

### DIFF
--- a/datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
+++ b/datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
@@ -28,7 +28,6 @@ import org.nd4j.linalg.api.memory.pointers.PagedPointer;
 import org.nd4j.linalg.api.ndarray.INDArray;
 import org.nd4j.linalg.exception.ND4JIllegalStateException;
 import org.nd4j.linalg.factory.Nd4j;
-import org.nd4j.linalg.primitives.Pair;
 import org.nd4j.linalg.util.ArrayUtil;
 
 import java.io.*;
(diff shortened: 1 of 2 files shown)


datavec-data/datavec-data-image/src/main/java/org/datavec/image/loader/NativeImageLoader.java
```diff
@@ -205,10 +205,16 @@ public INDArray asMatrix(File f) throws IOException {
...
     @Override
     public INDArray asMatrix(InputStream is) throws IOException {
-        byte[] bytes = IOUtils.toByteArray(is);
-        Mat image = imdecode(new Mat(bytes), CV_LOAD_IMAGE_ANYDEPTH | CV_LOAD_IMAGE_ANYCOLOR);
+        Pair<byte[],Integer> pair = streamToBuffer(is);
+        byte[] bytes = pair.getFirst();
+        int length = pair.getSecond();
+        BytePointer bp = new BytePointer(length);
+        bp.put(bytes, 0, length);
+        Mat mat = new Mat(bp, false);
```
saudet: We don't need to create a BytePointer explicitly here, just call `new Mat(bytes)` like before and it *should* work.

BTW, creating a new BytePointer like that on each call is also bound to create a lot of garbage. You're not seeing that?

saudet: Let me take a stab at it

AlexDBlack:
@saudet Any progress here? Or should we merge this as is (to reduce the byte[] memory use) and look at the pointer/offheap component separately?

saudet:
Ah yes let me do this now actually.

PR #580, opened 2018-05-13
add timing stats module

agibsonccc:
Adds InputStreamInputSplit support for image record reader and IOTiming module.

## How was this patch tested?

Please review


datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/FlipImageTransform.java
```diff
@@ -80,7 +80,12 @@ protected ImageWritable doTransform(ImageWritable image, Random random) {
         if (image == null) {
             return null;
         }
+
         Mat mat = converter.convert(image.getFrame());
+
+        if(mat == null) {
+            return null;
```
AlexDBlack: I'd double check with @saudet here, but that seems like it would cause issues in the next transform?

datavec-perf/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+        <groupId>org.datavec</groupId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>datavec-perf</artifactId>
+
+    <name>datavec-perf</name>
+    <!-- FIXME change it to the project's website -->
+    <url>http://www.example.com</url>
```
AlexDBlack: Fix/delete URL

datavec-perf/pom.xml
```diff
@@ -0,0 +1,85 @@
...
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <version>4.11</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <pluginManagement><!-- lock down plugins versions to avoid using Maven defaults (may be moved to parent pom) -->
```
AlexDBlack: Plugin management section not necessary?
And we have versions already specified in parent pom...

datavec-data/datavec-data-image/src/main/java/org/datavec/image/transform/FlipImageTransform.java
agibsonccc: Ahh, I think that might have been an artifact from a previous commit. That being said, what should we do when a transform fails is a broader question I think we should answer here.

datavec-perf/pom.xml
agibsonccc: Addressed. That was just boilerplate I forgot to remove.

